PR #12, opened 2014-06-15
Merge allgress query analysis branch to tonsky.

allgress:
This branch adds query analysis and the ability to listen only for transactions that would affect a particular query. The query analysis remembers which parts of the index are used in the query, and transacted datoms are analyzed to see if they update any of these. May be a useful performance optimization, rather than notifying all listeners of all transactions and leaving it to the listener to query against tx-data, decide if any changes are relevant, etc.

Not sure if this is the best approach, particularly since the logic is essentially duplicated between the query and analysis functions. Also some annoying whitespace changes I didn't catch before committing, sorry.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -5,12 +5,12 @@
   :license {:name "Eclipse"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
   :url "https://github.com/tonsky/datascript"
-  
+
   :dependencies [
     [org.clojure/clojure "1.6.0"]
     [org.clojure/clojurescript "0.0-2227"]
   ]
-  :cljsbuild { 
+  :cljsbuild {
     :builds [
       { :id "release"
         :source-paths ["src"]
@@ -27,7 +27,7 @@
         [lein-cljsbuild "1.0.3"]
         [com.cemerick/clojurescript.test "0.3.0"]
       ]
-      :cljsbuild { 
+      :cljsbuild {
         :builds [
           { :id "dev"
             :source-paths ["src" "test"]
--- a/src/datascript.cljs
+++ b/src/datascript.cljs
@@ -1,15 +1,15 @@
 (ns datascript
   (:require
-    [clojure.set :as set]
-    [clojure.walk :as walk]
-    [cljs.reader :refer [read-string]]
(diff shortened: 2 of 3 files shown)


kahunamoore:
This sounds a lot like what the RETE algorithm does - it effectively builds a custom network/indexes that exactly fit the set of queries/rules known to the system and shares common sub expressions for efficiency. Trades memory for runtime.

jeluard:
Are you still considering this feature? Looks extremely useful.

tonsky:
@jeluard yes, but haven’t got time to implement it or do a review on PR. At some point, something like this, yes

atroche:
@allgress are you guys still using this for your own apps? how is it working for you?

sparkofreason:
@atroche We're using Datascript, but not the allgress branch with query-based notifications. We've moved away from the single state atom approach, and have really embraced the immutable aspect of Datascript. Our web components might receive an instance of a Datascript DB via an attribute from a parent, or create it themselves through a service call. Child components often publish relevant changes through events, from which listeners can update their own instance as appropriate. Since we use freactive, sometimes we'll make a lens which represents a specific query or "view" on the DB.

We've found that this approach allows us to keep the Datascript instances fairly small (for our apps, at least), and haven't had any perf issues which would require query-specific notifications.

kahunamoore:
No, we are using the Clara rules engine which gives you these features (and more) by design.

metasoarous:
Hadn't seen this thread before... I'll just chime in to mention that this sounds a lot like parts of https://github.com/mpdairy/posh. In fact, there's an `analyze-q` there as well with output results that look a lot like these, so I'm wondering if @mpdairy actually looked at that in his implementation. To my knowledge he hasn't hooked any of that up for generalized listeners, but there's a single listener that goes through the active subscriptions, checks datoms against the generated patterns, and decides whether or not to update the queries based on this. So, if anyone reading this thread is interested in this line of thinking, you may want to check out Posh.

That having been said, I'd like to point out that there's some discussion on #132 in the direction of incremental materialized view maintenance which would go one step further in actually preventing re-computation of queries, instead using previous results and tx-diffs to compute the new query results far more efficiently.

Issue #58, opened 2015-02-16
Infinite loop when using pull API on circular component reference


cigitia:
In DataScript 0.9.0, as far as I can tell, the `pull` function may lock the program in an infinite loop if it is used on an entity whose component attribute refers to itself, in a circular reference.

A demonstration script is at
https://gist.github.com/cigitia/c1f1fc48ebc041c979c5.
Line 24 needs to be uncommented to freeze the program.

This especially may happen when creating components in the same transaction as their parents, due to current behavior in the generation of component entities’ IDs; see #59.

DataScript is amazing; thank you for creating it.

tonsky:
Will take a look at this a little bit later. If I remember correctly, components fetch is unconditional, so if they are in a cycle, it’s an infinite loop. Datomic hangs on this too. Idea is to not put your components in a cycle. Probably there’s no reason to emulate that behaviour of Datomic here.

cigitia:
Thanks for considering it; if Datomic also freezes on this, thinking more about whether to check for this is understandable. There are three options that I can think of:

* Leave the infinite-loop behavior completely in; no checks at all.
* Use `assert` to check for infinite loops to throw errors before they occur; can be enabled or disabled for different compiler builds using the [`:elide-asserts` option](https://github.com/clojure/clojurescript/wiki/Compiler-Options#elide-asserts); leaves in infinite loops only on optimized builds for which `:elide-asserts` is `true`.
* Check for infinite loops and terminate them, causing circular references to be resolved without throwing errors or infinitely realizing more objects.

In addition, the touch API probably also suffers from this same problem since, when used on an entity, it recursively realizes its components too; I haven't yet tested this, though.

tonsky:
I think there’s no reason for allowing infinite recursion for components. I guess Datomic does that is a bug. So we’ll fix that eventually, @dthume maybe you can take a look?

tonsky:
Fixed in #61 and #62

PR #61, opened 2015-02-25
First pass at fix for #58

dthume:
This PR provides similar behaviour to recursion specs (i.e. only returns a map containing `:db/id` for entities which have already been seen along a given path). I have to be honest, though, and say that I'm not convinced allowing loops on component attributes is the right way to go; `:db/isComponent` currently allows one to explicitly constrain parts of the schema more strictly than just `:db/ref` - in the pull API this is used to, for example, make the assumption that navigating a _reverse_ component relationship will yield precisely 0 or 1 results depending on whether the attribute exists on the given entity.

It also seems to violate the principle of least surprise (although granted, the original issue that caused this was clearly surprising, but it was caused by a separate issue) - To my mind `:db/isComponent` conveys information on multiple levels:

* On a schema level it indicates an explicit choice of a "contains" relationship over the less constrained, plain `:db/rel`.
* `transact` etc. use it to supported nested maps in transaction data.
* `db/retractEntity` uses it to make deletion decisions that are different from those of plain `:db/rel` based upon this constraint.
*  `touch` / `pull` / `pull-many` etc. use it to make expansion decisions based upon the implied contains constraint that are different from those of plain `:db/rel` - the single parent requirement, for example.

The latter items all seem to build upon the constraint provided by the first item - it therefore seems dangerous to remove the constraint itself.

For example this PR exhibits some rather odd behaviour if you give it graphs with multiple component attributes containing mutual cycles - see the commented out [test](https://github.com/dthume/datascript/blob/pull-component-recursion/test/datascript/test/pull_api.cljs#L130

On the one hand, it appears to be "technically" correct from the point of view of not returning entities which have already been seen along a given path, but on the other it doesn't seem particularly easy to work with.

All that said, perhaps I'm just overthinking things; if you decide you'd like to proceed with the PR, by all means let me know if there's anything you'd like changed.

### DIFF
--- a/src/datascript/pull_api.cljs
+++ b/src/datascript/pull_api.cljs
@@ -85,8 +85,12 @@
 
 (let [pattern  (dpp/PullSpec. true {})]
   (defn- expand-frame
-    [attr-key multi? eids]
-    (subpattern-frame pattern eids multi? attr-key)))
+    [parent eid attr-key multi? eids]
+    (let [rec (-> (:recursion parent)
+                  (update attr-key push-recursion eid))]
+      (-> pattern
+          (subpattern-frame eids multi? attr-key)
+          (assoc :recursion rec)))))
 
 (defn- pull-attr-datoms
   [db attr-key attr eid forward? datoms opts [parent & frames]]
@@ -111,10 +115,10 @@
                         (mapv datom-val found)
                         eid parent frames)
 
-          (and forward? component?)
+          (and component? forward?)
           (->> found
                (mapv datom-val)
-               (expand-frame attr-key multi?)
+               (expand-frame parent eid attr-key multi?)
                (conj frames parent))
           
           :else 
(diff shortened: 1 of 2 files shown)


tonsky:
Well, I totally agree there’re an implied constraints about components. Unfortunately, it’s hard to actually enforce it: there’re actually a lot of indirect ways to create loops, and whole DB should be checked on each transaction that no new violations have been introduced.

Probably nobody will ever find it out, if components will be uses as we expect them to be used. But just in case, it’s good that system does not fall apart if such problems will be introduced by someone.

Again, in normal usage these situations should not ever happen. So this is just some sort of defensive mechanism, just in case.

Thanks for the PR, I’ll look into it shortly

tonsky:
Just to be more clear: we work in expectation that all these constraints holds true. E.g. reverse ref navigation from component returns single result. User _can_ make two entities to reference single component, but then it’s his problem to deal with it.

tonsky:
Merge this for now so it’s in master, needs more work — see #62

PR #65, opened 2015-03-20
Expose Filtered DBs API to js

ActiveObject:
I was looking around and found that filter function isn't available in js version.


tonsky:
thanks for pointing that out!

Issue #45, opened 2014-12-22
Can't figure out how to empty database with listeners


wambat:
Can't figure out how to "clean" database with listeners attached. 
I don't want to go through retract/add on entities, just clean it, and make all listeners jingle.
That's seems not very idiomatic, but how would one practically reset database?
PS great lib, thanks, fills a huge gap for me.

cigitia:
It’ll be wonderful when this gets added. When it does get implemented, it might be best to implement it as a transaction function that can be used in `transact!`, like `:db/add` and `:db.fn/cas`, rather than a function separate from `transact!` that acts directly on connections.

tonsky:
@cigitia yep, sounds reasonable

tonsky:
Just added `reset-conn!` function to reset database value inside a conn and trigger all listeners. Usage example: https://github.com/tonsky/datascript/blob/master/test/datascript/test/conn.cljc

@cigitia I would pass on your idea for now, if you need that kind of function inside a transaction it’s not that hard to build for youself. Transactions totally support user-provided functions

PR #67, opened 2015-04-01
Allow js api to create conn from db

sherbondy:
Maybe I am going about this the wrong way,
but I am trying to deserialize a datascript database
from localstorage (a la your example here: https://github.com/tonsky/datascript-todo/blob/gh-pages/src/datascript_todo.cljs)
but via the vanilla js api.

There seems to be an asymmetry in the js api,
in that you can dereference a database from a conn,
but you cannot easily create a conn from a database.

My pull request is kind of a dumb hack.
Maybe it would be better for the original create-conn
function to be a multimethod or something.

### DIFF
--- a/src/datascript/js.cljs
+++ b/src/datascript/js.cljs
@@ -94,6 +94,10 @@
 (defn ^:export create_conn [& [schema]]
   (d/create-conn (schema->clj schema)))
 
+(defn ^:export conn_from [db]
+  (atom db 
+        :meta { :listeners  (atom {}) }))
+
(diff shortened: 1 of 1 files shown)


tonsky:
Yep, that should be fixed. I’ll address that in the next version

sherbondy:
Hey, you know what, my reset_db suggestion is essentially a repeat of this issue:
https://github.com/tonsky/datascript/issues/45

So I'll close this issue, and will open a new one if #45 gets resolved without being exported to JS-land :)

Issue #54, opened 2015-02-04
Clojure version?


jeroenvandijk:
I have a special use case where I (probably) don't want to use the in-memory Datomic version [1], but I do want all the Datomic features available in an in-memory fashion, just like Datascript. So a few questions:

* Would you like to have a Clojure version too? Does it make sense to you?
* Do you think it would require a lot of changes to the current codebase?

[1] It has unnecessary dependencies (when running in-memory). And I'm not sure about the properties of the in-memory version for long running processes and I can not easily find out due to the unknowns about the internals of Datomic.

tonsky:
Eventually there will be a Clojure version. Unfortunately it’s not a straightforward port, mostly because performance optimizations are platform-specfic. I’m planning on doing this after all main features of Datomic API will be implemented in DataScript, so I’ll avoid developing 2 implementations at the same time.

jeroenvandijk:
Thanks for the quick response. I asked on the Datomic mailinglist [1] and I'm more convinced now that a Clojure DataScript provides new use cases. I'll try with Datomic first, but I'm might go the DataScript route if it turns out to difficult or to hacky. 

[1] https://groups.google.com/forum/#!topic/datomic/Cq7ULo0sSBc

hansgru:
+1 for a Clojure version

jeroenvandijk:
FYI, I made a start https://github.com/tonsky/datascript/compare/master...jeroenvandijk:feature/clojure-port. I try to get the test suite over to cljx, step by step, until I have both `cljsbuild test` and `test` passing via:
```
lein do clean, cljx once, cljsbuild test, test
```
Not sure how much time it will take me yet. I've copied the cljx approach from Prismatic Schema. Does this sound ok to you?

tonsky:
Well, test suite can almost transparently being copied to clj—there’s nothing cljs-specific in tests or datascript api at all. Other stuff will need more creative approach: e.g., btset needs a complete rewrite, query partially depends on #js stuff, debug uses a lot of cljs-dependent things. Plus there’re protocol differences in base clojure/clojurescript classes. You can experiment with this to identify possible problems and solutions, but I’m not sure I’ll be able to merge your branch as-is, a lot is changing right now, and merge will probably be a disaster.

jeroenvandijk:
That makes sense. I'll regard the first as an experiment. Thanks
Op 28 feb. 2015 09:32 schreef "Nikita Prokopov" <notifications@github.com>:

[...]

PR #68, opened 2015-04-09
clojure port

benfleis:
This represents a solid beginning toward a clojure port, see https://github.com/tonsky/datascript/issues/54.

Lots of tweaking done, and more is needed. Inlined/XXX comments explain reasoning or open questions. Best place to actually add comments is the second commit: https://github.com/benfleis/datascript/commit/d943eaba47da49cf8d647f1eb90068ed31f14644. (The first commit was only file movement, to make diff viewing easier.)

What would it take to turn this into a basis for a real clojure/script version of datascript? I suspect that the big consideration points would be:

* value vs. nuisance
* portable code, performant code
* cljx/cljc (1.7 reader conditionals)

It runs all CLJ tests, although the performance suite is not fully supported. The clojurescript code does not work at the moment, although that will be resuscitated soon. (I decided at a certain point on just getting the clojure side to work.) The fork originates from early february, so there would be a fair bit of catching up to do as well.

Feedback, thoughts, suggestions?


benfleis:
All of this reworked and ultimately incorporated in, via a series of pulls:

  https://github.com/tonsky/datascript/pull/70
  https://github.com/tonsky/datascript/pull/71
  https://github.com/tonsky/datascript/pull/78
  https://github.com/tonsky/datascript/pull/79
  https://github.com/tonsky/datascript/pull/82

PR #70, opened 2015-04-16
move nearly all cljs files to cljc, as basis for work on clojure

benfleis:
port. updated compilers: clj -> 1.7.0-beta1, cljs -> 0.3196


jeroenvandijk:
Nice!

PR #71, opened 2015-04-17
import/adjust btset.clj, update test/.../btset.cljc to run in both

benfleis:
2 things worth bringing up:
- formatting the #?@(:cljs [ ... ]) -- i left defns fully on the RHS. thoughts?
- the hash-map change at the bottom of the test code is to deal with a known bug in clj 1.7.0-beta1 compiler: http://dev.clojure.org/jira/browse/CLJ-1698

thoughts?

### DIFF
--- a/src/datascript.cljc
+++ b/src/datascript.cljc
@@ -251,15 +251,24 @@
       :else   s)))
 
 (defn squuid []
-  (UUID.
-    (str
-          (-> (js/Date.) (.getTime) (/ 1000) (Math/round) (to-hex-string 8))
-      "-" (-> (rand-bits 16) (to-hex-string 4))
-      "-" (-> (rand-bits 16) (bit-and 0x0FFF) (bit-or 0x4000) (to-hex-string 4))
-      "-" (-> (rand-bits 16) (bit-and 0x3FFF) (bit-or 0x8000) (to-hex-string 4))
-      "-" (-> (rand-bits 16) (to-hex-string 4))
-          (-> (rand-bits 16) (to-hex-string 4))
-          (-> (rand-bits 16) (to-hex-string 4)))))
+  #?(:clj
+      (let [uuid     (java.util.UUID/randomUUID)
+            time     (int (/ (System/currentTimeMillis) 1000))
+            high     (.getMostSignificantBits uuid)
+            low      (.getLeastSignificantBits uuid)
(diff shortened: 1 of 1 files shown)


benfleis:
@tonsky damnit. will withdraw while I figure out what i've done wrong here.

tonsky:
Nice! Our first real step towards CLJ version :)

- Can you make `semantic-test-slice` work with CLJS too? If we’re aiming at impl transparency, tests should be the same for both imls.
- I don’t get why are you testing `(sorted-set-by cmp-s)`, not `btset`. I understand that it’s just a wrapped sorted-set, but still, we should test around public interface, not impl detail. Correct me if I miss something.
- I think helpers at the beginning can be safely dropped (decode-path, dump)
- These
```
(defn btset-conj [set key] (conj set key))
(defn btset-disj [set key] (disj set key))
```
should read
```
(defn btset-conj [set key _] (conj set key))
(defn btset-disj [set key _] (disj set key))
```
right?

Thanks!

benfleis:
Everything sounds good. 2 things worth telling:

> I don’t get why are you testing (sorted-set-by cmp-s), not btset. I understand that it’s just a wrapped sorted-set, but still, we should test around public interface, not impl detail. Correct me if I miss something.

You are correct - I originally wrote it since the first btset version I wrote w/ sorted-set-by passed the existing tests, but was actually broken, so I went back a step and wrote the sorted-by-set tests to make certain it was correct. I will keep the tests, but w/ the btset API. Should have done that in the first place, but didn't think much about it from that POV.


```
(defn btset-conj [set key] (conj set key))
(defn btset-disj [set key] (disj set key))

 ->

(defn btset-conj [set key _] (conj set key))
(defn btset-disj [set key _] (disj set key))
```

I debated this, and in the end thought it better to have it fail when compiling in CLJ, since it feels misleading to accept and ignore that argument, since it would have a clear semantic expectation attached to it. I thought the alternative would be to leave the argument, with an assert failing any time it's not nil. Whatcha think?

tonsky:
`btset-conj` with 3 args was just a way to provide faster comparator where it made sense. Since we’re using stub impl for btset for now, it’s ok to ignore that optimization as well. `datascript.core` will still pass custom comparator there, but we can safely ignore that, all we lose is some perf. When we had proper btset impl, it will make use of it.

PR #74, opened 2015-05-01
review PR for next step of imports, but with many questions below.

benfleis:
port everything necessary to run (btset), core, and entity tests, which
implied "full" import of datascript.cljc, ds/{core,macro,parser}.cljc;
all other modified files are "minimal" to support continued
compilation/work.

My intention is to fix these things up, and merge all into a single
clean diff in the move-to-cljc branch.

Open questions before this is truly ready for merge submission:

- FilteredDB -> deftype?  (defrecord doesn't allow redefines w/o
  warning, but extend-type does - logical, but unclean) also wary of
  maintaining separate types, with likely different behaviors, no matter
  how hard we try to keep in sync

- some functionality moved from datascript -> datascript.core to avoid
  import/ns/object ugliness. could do it the other way and do more
  duplicate import/refer stuff with classes; it felt more natural this
  way, since it co-locates more related functions in one place.

- unclear how much to name w/ '-' prefix, unsure of intent - CLJS
  suggests that CLJS core things should be '-' prefixed, but others
  not. many of existing protocols use '-' prefix, some internal funcs
  do, many do not.

- macros got all put into macro.cljc, before I got this answer:
  http://stackoverflow.com/questions/29914967/can-cljc-single-file-macro-definitions-to-work-with-clojurescript
  debating whether to put macros back into their places of use (almost
  all are single file or single use)

- clojurescript.test suggests that :none optimization doesn't work with
  tests -- so i switched it to :whitespace (and renamed for
  disambiguation); thoughts?

- entity has assoc in order to support clojure.walk, which afaict may
  only be used in entity-map.

- touch API -- if datoms search comes back empty, shouldn't it still be
  marked as touched?

- formatting - feels like setting a precedent w/ the #?( ... ) - I tried
  to preserve "top level" things as such, as much as possible, open to
  suggestion.


tonsky:
So I read all the diff, here are my notes, in read order:

- Point of `deftype` on `Datom` was to avoid unnecessary fields (meta, extended attrs map) so it’s a lightweight as possible. For all other cases `defrecord` + `extend-type` works just fine and is cross-platrofm, so let’s stick to that. The parts where you’re trying to implement what `defrecord` natively implements are most fragile ones, so let’s try keep them to minimum. `val-at-db`, `assoc-db` and the likes should go

- `entryAt` and `equiv` are swapped in `Datom`/clj version

- Let’s move macros from `macros.clj` to `core.cljc` and `deftrecord` to `parser.cljc`, and remove `macros.clj`

- For partially implemented methods, we should throw `UnsupportedOperationException` in clj, this is idiomatic in Java. Also please remove `assoc-datom`/`empty-datom` and the likes, they clutter the namespace. It’s ok to just throw in-place, no need to generalize such task

- `entryAt` should return `clojure.lang.MapEntry`, not vector

- About `cmp-val`/`cmp-val-quick` formatting. I prefer to have one function header, and then conditional reader in the fn body. This way it reads: there’s one fn, but impl may differ. It’s easier to understand: ok, I know fn exist, always. On the contrary, the way you have it now reads: there’s some fns that work only for CLJS, and here are some fns that work only for CLJ. It’s a subtle difference, I know, but I prefer it my way.

- `cmp-val-quick`: note that `compare-keywords` is already gone, we resort to default compare now in CLJS too

- In `cmp-datoms-*-quick` indentation is broken (should be 4 spaces instead of 3)

- `get-datoms-as-tuples` will materialize vector of all datoms in memory, might be big. We can probably use lazy seqs here somehow

- `#?(:clj (do …))` → `#?@(:clj […])`

- What’s the point of `db*`? Also, I prefer `map->DB` for db creation as there’s a lot of keys in DB

- Why `empty-db` is overloaded? I think `arg` should always mean the same thing (`schema` in our case)

- Why removing `^boolean` annotation for `is-attr`/`multival`/… methods?

```
-    (not (or (array? vs)
+    (not (or (#?(:cljs array? :clj vector?) vs)
```

we have `array?` defined in `datascript.core` earlier, no need for conditional read here

- `(assert db)` — probably forgot to remove

- `retract-components`: you replaced transducer form with lazy one, why? It’s less performant that way

## PARSER

```
-    (when-not (empty? undeclared-vars)
+    (when (seq undeclared-vars)
```

`empty?` is allocation-free and in most cases just and integer check. `seq` creates a new object if sequence is not empty (`ChunkedSeq` for vector, `KeySeq` for set, etc). Also there’re optimizations (in CLJS at least) for conditions which can be proved to have boolean type. In case of `seq` it has to call additional `truth_` function to coerce to boolean. So please don’t do this

## QUERY

```
-  ([] (Relation. {} [#js[]]))
+  ([] (Relation. {} [#?(:cljs #js [])]))
```

This is not correct for CLJ version. There must be a single 0-tuple element in this rel.

```
-  (or (seqable? x) (array? x)))
+  (or (dc/seqable? x) #?(:cljs (array? x))))
```

again, `dc/array?` to the resque

- Is it possible to keep types from parser namespace-qualified? I prefer not to use `use`/`refer` until absolutely necessary.

## QUERY_V3

```
-  (or (seqable? x) (array? x)))
+  (or (dc/seqable? x) (array? x)))
```

needs `dc/array?` too

## ENTITY

- How come `clojure.walk` needs `assoc`?

- What’s `TouchableEntity?` Why do we need this?

- We used `set!` for `Entity` to keep mutable cache. It can be changed to `volatile!` ref, so impl can be shared between clj and cljs. I very much prefer it is shared

## TESTS

- Does clojure.test fail to print ExceptionInfo data too? If it does print it, we probably doesn’t have to patch test failure printer.

```
-      (is (= (-> (e 10) :children first :father) (e 10)))
+      (is (some #(= (e 10) (:father %)) (-> (e 10) :children)))
```

It’s strange you had to change that test. Can you explain why?

## Everything else

Formatting: looking at conditional reader forms, it’s hard to read now. My proposal: let’s treat them as any other form. So top-level forms included into `#?()` will be indented, but I think it’s ok. Like this:

```
#?(:clj
  (defn array? [o]
    (-> o type .isArray)))
```

> clojurescript.test suggests that :none optimization doesn’t work with tests — so i switched it to :whitespace (and renamed for disambiguation); thoughts?

How do you run tests? I open dev.html in Chrome and connect with REPL, works perfectly with any optimization level.

> unclear how much to name w/ ‘-‘ prefix

I’m ok to drop it for fns, but keep for protocols. Please don’t do this over the whole codebase, only the fns you touch.

> some functionality moved from datascript -> datascript.core

Sure, it’s absolutely logical.

> touch API — if datoms search comes back empty, shouldn’t it still be marked as touched?

Doesn’t really matter, `touch` is for interactive exploration anyway. There’s no place for it in a working program.

> I wonder what you’d think of a different form — adding another clause (DS specific, ignored by Datomic, at least in {} form), like :binding or :function, which could itself be a map or sequence of symbol, and resolved function. This would give a language that works identically in CLJS, and could be ignored/dropped in CLJ, since resolve can actually be called there.

Good idea. Will check out how to make it so Datomic ignores these.

> My intention is to fix these things up, and merge all into a single clean diff in the move-to-cljc branch.

This is kinda hard to read. It took me an hour to speed-read, and I’m sure I miss a lot of important nuances. Diff is very big and it mixes a lot of things. As a result, while I’m not 100% sure of _everything_ is this diff, I cannot merge _anything_ from it. Another downsides:

- I should re-read _everything_ on every iteration, even stuff I already reviewed and we both agree on.
- There’s no feeling of moving forward
- Patch rots. I commit to master constantly, and you’ll spend a lot of time rebasing. It’s tedious, and you can easily miss something (e.g. compare-keywords).

That’s why I prefer small steps and clean small separate diffs. Not put everything in one “clean” patch (it’s a long road until it’s clean), but separate patches by logically independent pieces. E.g. `(Record.)` → `(->Record)` migration, it’s easy to review and merge if diff contains only this.

I understand that you’ve tried to get to compiled code and working tests first and these are the minimal changes needed to do this. But it doesn’t mean patch has to be one giant diff between master and this.

Also, it’s important to me to have some confidence in the code I merge, so it’s hard to accept patch that somehow works and is intended to be fixed later. I need it to be incomplete, but “right” from the beginning.

benfleis:
## Big Picture

> That's why I prefer small steps and clean small separate diffs. Not put everything in one “clean” patch (it's a long road until it's clean), but separate patches by logically independent pieces. E.g. `(Record.)` → `(->Record)` migration, it's easy to review and merge if diff contains only this.

I'm on board. Will split out into much smaller chunks.


## Specific responses

> So I read all the diff, here are my notes, in read order:
> 
> - Point of `deftype` on `Datom` was to avoid unnecessary fields (meta, extended attrs map) so its a lightweight as possible. For all other cases `defrecord` + `extend-type` works just fine and is cross-platrofm, so let's stick to that. The parts where you're trying to implement what `defrecord` natively implements are most fragile ones, so let's try keep them to minimum. `val-at-db`, `assoc-db` and the likes should go

Completely agreed. The factor here is that equivalence and hash values cannot be overridden with a defrecord in CLJ. Initially, I only changed it when necessary for this reason. Afterward I believe I switched at least one CLJS defrecord -> deftype for the sake of behavioral consistency between the versions. 

If extend-type allows it in CLJ, then I will gladly revert -- I prefer taking what the language(s) give us, as much as possible. (I don't recall any longer whether I tried redefining equiv/hash in extend-type in the first place.) 

> - `entryAt` and `equiv` are swapped in `Datom`/clj version
> - Let's move macros from `macros.clj` to `core.cljc` and `deftrecord` to `parser.cljc`, and remove `macros.clj`
> - For partially implemented methods, we should throw `UnsupportedOperationException` in clj...
> - `entryAt` should return `clojure.lang.MapEntry`, not vector

Check.


> - About `cmp-val`/`cmp-val-quick` formatting. I prefer to have one function header, and then conditional reader in the fn body...

Cool -- I generally prefer it that way too, but gave both perspectives so you can see and compare.

Re cmp funcs, easy fix.

 
> - `get-datoms-as-tuples` will materialize vector of all datoms in memory, might be big. We can probably use lazy seqs here somehow

I will have a look -- I remember considering this when I originally wrote it, but I am no expert on lazy, or perhaps lost my intention.


> - `#?(:clj (do …))` → `#?@(:clj […])`

This is not allowed in top level forms. There's a bug filed in CLJ someplace about it. It took me a while to discover this though.

  http://dev.clojure.org/jira/browse/CLJ-1706



> - What's the point of `db*`? Also, I prefer `map->DB` for db creation as there's a lot of keys in DB

Originally to declare. Will clean this up.  `map->DB` doesn't exist if it's deftype.


> - Why `empty-db` is overloaded? I think `arg` should always mean the same thing (`schema` in our case)

Not smart -- original thinking was (empty foo) -> (empty-db db) consistency, but that could be privatized into a * variant.

 
> - Why removing `^boolean` annotation for `is-attr`/`multival`/… methods?

It caused test failures in CLJ world; iirc, it's because the funcs don't actually return bools, but instead truthy things. Could also have wrapped return w/ (boolean). 


> we have `array?` defined in `datascript.core` earlier, no need for conditional read here

Yep, that was written before I wrapped it, needs fix.

 
> - `(assert db)` — probably forgot to remove
> 
> - `retract-components`: you replaced transducer form with lazy one, why? It's less performant that way

Also originally written versus 1.6, all of those should be reverted (there are several).

Other than Entity, I'll skip the rest, since we'll move to a more per-file thing. I will come back to refer to these when it's time.

> ### PARSER
> 
> `empty?` is allocation-free and in most cases just and integer check. ... So please don't do this

Check.


> ### QUERY
> 
> ```
> -  ([] (Relation. {} [#js[]]))
> +  ([] (Relation. {} [#?(:cljs #js [])]))
> ```
> 
> This is not correct for CLJ version. There must be a single 0-tuple element in this rel.

This should have had a `; XXX incorrect for CLJ after it, as a marker. not fully imported.`


> ### ENTITY
> 
> - How come `clojure.walk` needs `assoc`?

  https://github.com/clojure/clojure/blob/2224dbad5534ff23d3653b07d9dc0a60ba076dd7/src/clj/clojure/walk.clj#L49

`(into (empty <defrecord>))` turns into a series of conjs. Since it's essentially a keyed structure, these conjs become assocs.


> - What's `TouchableEntity?` Why do we need this?

I no longer recall why I made it a separate protocol. Probably no good reason :)

 
> - We used `set!` for `Entity` to keep mutable cache. It can be changed to `volatile!` ref, so impl can be shared between clj and cljs. I very much prefer it is shared

I spent a bit of time trying to get `:volatile-mutable` working with deftype. I failed. I asked about it on #clojure, and nobody seemed to even know of a real example, or how my example work. So I gave up. I'll give it another shot. 
 
> ### TESTS
> 
> - Does clojure.test fail to print ExceptionInfo data too? If it does print it, we probably doesn't have to patch test failure printer.

Ahh, interesting. I literally just ported it blindly, without testing it directly. Probably not needed.


> ```
> -      (is (= (-> (e 10) :children first :father) (e 10)))
> +      (is (some #(= (e 10) (:father %)) (-> (e 10) :children)))
> ```
> 
> It's strange you had to change that test. Can you explain why?

The original test was dependent upon ordering within the set, which didn't work in CLJ.


> ### Everything else
> 
> Formatting: looking at conditional reader forms, it's hard to read now. My proposal: let's treat them as any other form. So top-level forms included into `#?()` will be indented, but I think it's ok. Like this:

Easy peasy.


> > clojurescript.test suggests that :none optimization doesn't work with tests — so i switched it to :whitespace (and renamed for disambiguation); thoughts?
> 
> How do you run tests? I open dev.html in Chrome and connect with REPL, works perfectly with any optimization level.

I'd been running with :none as well, but switched to :whitespace when I saw that. I figured if the author says it doesn't work, then it's unwise to ignore. 


> > touch API — if datoms search comes back empty, shouldn't it still be marked as touched?
> 
> Doesn't really matter, `touch` is for interactive exploration anyway. There's no place for it in a working program.

True. You still prefer to leave it as is?

tonsky:
> The factor here is that equivalence and hash values cannot be overridden with a defrecord in CLJ

Ah, shit :( I’ll look for a better solution for this

> This is not allowed in top level forms. There's a bug filed in CLJ someplace about it. It took me a while to discover this though.

There was another bug about it, and it was fixed in beta2. Looks like they still have to work it out :(

> It caused test failures in CLJ world; iirc, it's because the funcs don't actually return bools, but instead truthy things. Could also have wrapped return w/ (boolean).

Yep, it’s better to make them actually return bools.

> (into (empty <defrecord>)) turns into a series of conjs. Since it's essentially a keyed structure, these conjs become assocs.

Datomic does not implement assoc/empty on them. I think we shouldn’t too. It just doesn’t make sense. `datascript.test.core` could be rewritten without it (or maybe change postwalk to prewalk?)

>> We used set! for Entity to keep mutable cache. It can be changed to volatile! ref, so impl can be shared between clj and cljs. I very much prefer it is shared

> I spent a bit of time trying to get :volatile-mutable working with deftype. I failed. I asked about it on #clojure, and nobody seemed to even know of a real example, or how my example work. So I gave up. I'll give it another shot.

No, I mean, there’s new ref type: volatile (e.g. `(volatile! 7)`), it’s like atom (`vswap!`, `vreset!`), but faster. Used internally by reducers. It works the same in clj and cljs.

> The original test was dependent upon ordering within the set, which didn't work in CLJ.

Oh, it’s probably because entity id 101 does not declare father attribute. We should fix that instead: 

```
(d/db-with
  [{:db/id 1, :children [10]}
   {:db/id 10, :father 1, :children [100 101]}
   {:db/id 100, :father 10}
   {:db/id 100, :father 10}]))
```

>  True. You still prefer to leave it as is?

Yes.

benfleis:
Everything in here got merged in, one way or another. Closing the pull, and removing the branch.

PR #75, opened 2015-05-03
Return nil from entity when passed nil eid

dwwoelfel:
Fixes a problem where calling (d/entity db nil) will throw an "Expected number or lookup ref for entity id" error.


tonsky:
Hope you don’t mind I didn’t used your patch :) Check out 0.11.1 version which has this issue fixed

PR #78, opened 2015-05-05
update entity code to function in CLJ

benfleis:
primary change is that both CLJ and CLJS use `volatile!`s instead of deftype `:mutable`-ish things to manage `touched` and `cache` fields. other changes accommodate common code tactics, with varying degrees of grace.

### DIFF
--- a/src/datascript/impl/entity.cljc
+++ b/src/datascript/impl/entity.cljc
@@ -1,8 +1,9 @@
 (ns datascript.impl.entity
-  (:require
-    [datascript.core :as dc]))
+  (:refer-clojure :exclude [keys get])
+  (:require [#?(:cljs cljs.core :clj clojure.core) :as c]
+            [datascript.core :as dc]))
 
-(declare Entity touch)
+(declare entity ->Entity equiv-entity lookup-entity touch)
 
 (defn- entid [db eid]
   (when (or (number? eid)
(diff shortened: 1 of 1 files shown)


tonsky:
- I’m thinking now: entities can be accessed from multiple threads. We probably should use atoms for cache instead of volatiles (volatile’s `vswap!` is safe in single thread only). Sorry I mislead you with my initial suggestion.
- Please set `(empty [e]         (entity (.-db e) (.-eid e)))` to throw
- We probably need IFn implemented too?
- What about removing `clojure.walk/postwalk` from test? Can you do it in this patch too?

benfleis:
I only used vreset! to avoid the race condition; since it's a cache, the worst case of a race is that the same cache element(s) get loaded twice. But your call -- it's not too hard to switch to atom.

I missed `empty`, will fix that one. IFn, will add. I was originally driven by making unit tests succeed, so some of the API elements are probably missing, when never used in the tests. (Also {key,entry,value_set} are missing, for example.) I think I was more thorough with later structures, but this first one left me wondering how far to go, and I never followed up by asking.

As to walk, I was separating out the test stuff and trying to keep the blocks much smaller; I am happy to dive into the rabbit hole, but the first version of this patch is what happens when following dependencies :) My branch will certainly reflect the walk change as we go, so I'm not worried about missing it.

PR #79, opened 2015-05-06
Integrate Datom CLJ port

benfleis:
which moves Datom from defrecord to deftype. Also move all Datom related funcs into same area, including a couple from datascript.cljc. Finally replace all `dc/Datom.` with `dc/->Datom.`.

### DIFF
--- a/src/datascript.cljc
+++ b/src/datascript.cljc
@@ -179,7 +179,7 @@
 (defn db-from-reader [{:keys [schema datoms]}]
   (init-db (map (fn [[e a v tx]] (dc/Datom. e a v tx true)) datoms) schema))
 
-(cljs.reader/register-tag-parser! "datascript.core/Datom" dc/datom-from-reader)
+(cljs.reader/register-tag-parser! "datascript/Datom" dc/datom-from-reader)
 (cljs.reader/register-tag-parser! "datascript/DB"         db-from-reader)
 
 ;; Datomic compatibility layer
--- a/src/datascript.cljc
+++ b/src/datascript.cljc
@@ -13,6 +13,8 @@
 (def  entity de/entity)
 (defn entity-db [entity] (.-db entity))
 
+(def  datom dc/datom)
+
 (def  pull dp/pull)
 (def  pull-many dp/pull-many)
 (def  touch de/touch)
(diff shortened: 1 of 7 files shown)


benfleis:
@tonsky ok, need to port the macros too - i had the old file still in my working tree and didn't realize it. will have fix shortly.

benfleis:
@tonsky FYI, I looked at doing the proper macro migration, but that won't work until core.cljc is fully ready to compile in CLJ, meaning the pr-str* form must be wrapper, and perhaps others. Soon...

tonsky:
Nice!

Some comments :)

UnsupportedClassVersionError → UnsupportedOperationException

```
(entryAt [d k] (some->> (val-at-datom d k) (clojure.lang.APersistentMap k)))
```
there should be MapEntry, not APersistentMap

assoc is actually needed (we need to be able to `(assoc datom :e 18)` for example). It should throw on unexpected keys though (IllegalArgumentException)

```
(defn datom [e a v & [tx added]]
  (->Datom e a v (or tx tx0) (if (nil? added) true added)))
```
does `deftype` also declares `->Datom` fn?

`(defn- equiv-datom [^Datom d o]` → `(defn- equiv-datom [^Datom d ^Datom o]`

`(apply pr (map (partial get d) [:e :a :v :tx :added])))` this probably better be expanded, so there’s no allocations and no lazy sequences on printing. It’s fine for interactive println/debug stuff, but same method will be used for serializing datoms to EDN, and for logs probably.

```
      #?(:cljs
         (if (identical? t1 t2)
           (compare o1 o2)
           (garray/defaultCompare t1 t2))
         :clj
         (combine-cmp
          (compare (str t1) (str t2))
          (compare o1 o2))))
```
Here this dance around `identical?` is actually needed. For same-type values we can do fast `(identical? t1 t2)` check and move straight to comparing values. In your clj case you always convert class to string and then compare these strings. Also `getName` is ~40% faster on a class than `toString`

```
(if (identical? t1 t2)
  (compare o1 o2)
  (compare (.getName ^Class t1) (.getName ^Class t2))
```

tonsky:
Another thing: I managed to measure `(Datom.)` vs `(->Datom)` perf (second one is a fn call, first one is a direct `new Datom` JVM instruction. Difference is not that big (~5%), but still, as we already got `(Datom.)` everywhere, can you _not_ change it to `(->Datom)`, just leave it as-is? It will also make diffs cleaner. Sorry that I haven’t decided that earlier

benfleis:
@tonsky

The misnamed classes are easy to fix -- one was a completion error, and the other an obvious mental miss.

I debated whether to keep assoc w/ Datom -- I had it in my original patch; I will re-add it.

The map thing w/ string output -- I had fixed it in one place, and will do the same there.

I think I know why the identical? check got changed - I had initially tried to reduce it to a compare, which doesn't work w/ all of the types, and went straight to strings. I am clearly not keeping optimization as high in my head as you are used to, and will push further in that direction.

As to `Datom.`, within core.clj, it's trivial. It's in the other files that it makes a difference -- then we have to add an import for every place that it's wanted. Intuitively I would argue for cleanliness and better dev-env over 5% gains, but I lack context to judge how important this particular 5% is -- I assume that query speed is the place where performance is most critical, and wonder how this affects queries. What are your thoughts here? I would for now propose using `Datom.` in core.cljc, and `->Datom` outside. Whatcha think?

tonsky:
Ah, decisions. There’s no strong reasons neither for `Datom.` nor `->Datom`. In cases like this I prefer just not to change things, leave them as-is. I’m not ready to do a proper research on how bad that will affect perf. Maybe a lot. Maybe slightly. Maybe not even slightly. Why do this while we can just use best possible alternative out there without much hassle? Import is not _that_ hard, and we have pretty small codebase. Let’s just keep `Datom.` as it was before.

PR #82, opened 2015-05-15
cljc breakout of DB and FilteredDB structures

benfleis:
* couple small fixes to Datom
* uses new defrecord-extendable-* macros to leverage defrecord, but allow overriding builtins
* possible mechanism to combine macros could be copied from prismatic, not yet tried
* DB: previous (and current) versions have `(seq)` support which returns a sequence of the `datoms`, but assoc, etc., will manipulate the DB structure directly. is this what we want? I might suggest removing the specialized seq, count methods until moving completely to the possible sorted-datom-set vision as a whole.

I will make cleanups in this branch, and once you're down with all of them, merge them into move-to-cljc for a final (singular) clean merge.

### DIFF
--- a/src/datascript.cljc
+++ b/src/datascript.cljc
@@ -19,6 +19,10 @@
 (def  pull-many dp/pull-many)
 (def  touch de/touch)
 
+(def  datom? dc/datom?)
+(def  db? dc/db?)
+(def  filtered-db? dc/filtered-db?)
+
 (def ^:const tx0 dc/tx0)
 
 (defn attr->properties [k v]
@@ -155,32 +159,11 @@
 (defn unlisten! [conn key]
   (swap! (:listeners (meta conn)) dissoc key))
 
-
-(defn pr-db [db w opts]
-  (-write w "#datascript/DB {")
-  (-write w ":schema ")
-  (pr-writer (dc/-schema db) w opts)
-  (-write w ", :datoms ")
-  (pr-sequential-writer w
-    (fn [d w opts]
-      (pr-sequential-writer w pr-writer "[" " " "]" opts [(.-e d) (.-a d) (.-v d) (.-tx d)]))
-    "[" " " "]" opts (dc/-datoms db :eavt []))
-  (-write w "}"))
-
-(extend-protocol IPrintWithWriter
-  dc/DB
-  (-pr-writer [db w opts]
-    (pr-db db w opts))
-
-  dc/FilteredDB
-  (-pr-writer [db w opts]
-    (pr-db db w opts)))
-
 (defn db-from-reader [{:keys [schema datoms]}]
   (init-db (map (fn [[e a v tx]] (dc/Datom. e a v tx true)) datoms) schema))
 
 (cljs.reader/register-tag-parser! "datascript/Datom" dc/datom-from-reader)
-(cljs.reader/register-tag-parser! "datascript/DB"         db-from-reader)
+(cljs.reader/register-tag-parser! "datascript/DB"    db-from-reader)
 
 ;; Datomic compatibility layer
 
--- a/src/datascript.cljc
+++ b/src/datascript.cljc
@@ -19,84 +19,14 @@
 (def  pull-many dp/pull-many)
 (def  touch de/touch)
 
+(def  empty-db dc/empty-db)
+(def  init-db dc/init-db)
+
 (def  datom? dc/datom?)
 (def  db? dc/db?)
 (def  filtered-db? dc/filtered-db?)
 
 (def ^:const tx0 dc/tx0)
-
-(defn attr->properties [k v]
-  (cond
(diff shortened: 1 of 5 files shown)


tonsky:
Great!

- `val-at-db`, `assoc-db`: these should be provided by defrecord now
- I now see that you were right about moving `empty-db` to core. I’m sorry, can you do that and drop `empty-db-from-db`?
- We should probably use `hash-ordered-coll` for `hash-db` cljs version (was a bug before). Also I think we can change to `(.-eavt db)` there instead of `(-datoms db :eavt [])`
- Would `(defn ^Boolean` work for cljs? Boolean is java class.
- In clojure version of `defrecord-ext` it’s safe to use field names (e.g. just `eavt` instead of `(.-eavt this)`). I think it’s much cleaner, can you change to that? I’ll see what can be done for CLJS version about that.

Thanks!

tonsky:
About DB/seq/set/protocols stuff. Let’s leave implementations exactly as they are, with all current flaws. Let’s revisit that decision in a separate ticket, later. It has to move one side or another, but let’s focus on getting this work first

benfleis:
These are easy -- funny that I forgot to drop the assoc/etc. operators.  The Boolean thing worked, because it got `def`d above, just like with the exception types.

The field name thing can work in the `(:clj ,,,)` wrapping, so we can make that switch there.

I also thought, after writing the description that 'extendable' isn't right, but 'replaceable' is, since that what it actually allows.

As to `empty-db`, there is still a need for 2 variants -- in an old patch, i combined them into a single func w/ an overloaded arg. I either need to do that, *OR* have 2 related variants, perhaps `(empty-db ...)` which is public and conforms to the current public API, and `(empty-db* ...)`, which would fulfill the normal  `(-empty ...)` need.

Should/may I also move the other related -db funcs into core? (I think empty, init, and the validation funcs.) I feel like they belong there next to the rest of the DB code, and they're all single use.

And as to protocols, I agree -- it was more to point it out as something to think about.

benfleis:
Also realize later: hash-db needs to work s/ FilteredDB too, so the type hint should go away, and then `-datom` is the right thing. Or am I missing sth? (This was my original intention, just the type hint is wrong.)

And the empty variant thing -- that's actually only the case if we worry about the other members being initialized differently, so that's actually not quite true. Can be used from both by passing the schema in from the defrecord calls. I'll try to get a patch up later today.

tonsky:
Just realized `hasheq` and `hashCode` should be cached on JVM. You’re right about DB/FilteredDB portability. OTOH the whole impl is so trivial we could inline it, without externalizing to `hash-db` at all.

This is what I meant by reusing empty-db: `(-empty [db] (empty-db (.-schema db))`

About moving fns to the core. Let’s postpone that decision. I’ll see what is the best way to organize things is.

benfleis:
doh, I see your last comment just now after pushing. /me -> chat.

tonsky:
You’ve moved `max-eid (if (pos? len) (.-e (aget datoms (dec len))) 0)` in `let` in `init-db`, it was important where it was because I’ve used mutable `.sort` on datoms. Can you return it back and add some comment about that?

benfleis:
re: hash caching -- it would either need to acquire an atom or other referential thing, or become a deftype. thoughts?

benfleis:
Random thought -- should the hash include the schema? Since equiv covers it, not necessary, but still feels weird.

benfleis:
@tonsky btw, I think the last diff contains everything we've talked about, here & in chat, so it should be ready for review.

tonsky:
I’ve fixed max-eid myself, see last push. Otherwise, looks good.

tonsky:
I want to check that ^:Boolean thing, and also new defrecord that lets use local field names in cljs

tonsky:
`^Boolean` thing is definitely not working:
```
(defn ^boolean test-aaa [x]
  (== x 1111))

(defn ^Boolean test-bbb [x]
  (== x 2222))

  (if (test-aaa 0)
    (println "aaa"))
  (if (test-bbb 0)
    (println "bbb"))


datascript.core.test_aaa = (function datascript$core$test_aaa(x){
return (x === (1111));
});

datascript.core.test_bbb = (function datascript$core$test_bbb(x){
return (x === (2222));
});

if(datascript.core.test_aaa.call(null,(0))){
cljs.core.println.call(null,"aaa");
}

if(cljs.core.truth_(datascript.core.test_bbb.call(null,(0)))){
cljs.core.println.call(null,"bbb");
}
```

Can you revert it to something like this (I can confirm it does work):
```
(defn #?@(:clj  [^Boolean test-ddd] 
          :cljs [^boolean test-ddd]) [x]
  "abc")
```

benfleis:
Boolean update is in place. As to the defrecord mods -- I assume you mean to do it by only emitting 'dupcliate' code in extend-type? Or just to make a similar macro for cljs side, so that the output is the result of defrecord, across the board?

PR #83, opened 2015-05-18
Compile under cljsbuild 1.0.6

tonsky:
- track clojure-1.7.0-beta3
- get rid of macro-only .clj files, moving their content into core.cljc
- get entity tests to work, implying:
    - update `ns` blocks in src, and a little bit in test
    - apply a number of small fixes from previous patch, mostly cljs -> clj tweaks
- add serialization tests, fix related code
- add a couple type hints to Datom/DB funcs, as a basis for reducing reflection warnings
- big progress toward getting everything to at least compile in CLJ, and be able to use cljsbuild 1.0.6

- open questions:
    - is-filtered predicate: consistency w/ other added type check predicates? (replace w/ filtered-db?)
    - could switch deftrecord to use defrecord-updatable, allowing specific instances to override, naturally (ITraversable, Not Or Variable)
    - ICache protocol unused. intentional?


tonsky:
- I this `is-filtered` is sufficient (name comes from Datomic), let’s remove `filtered-db?`
- I think `db?` should return true for both `DB` and `FilteredDB`

tonsky:
```
   (set! *data-readers* (merge *data-readers* '{datascript.core/Datom datom-from-reader
                                                datascript.core/DB    db-from-reader})))
```
`set!` seems to work only inside `(binding)` form. Let’s just provide `data_readers.clj` for Clojure case. CLJS branch of this is fine.

tonsky:
```
(defn- resolve-datom [db e a v t]
-  (when a (validate-attr a))
+  (when a (validate-attr a [e a v t]))
```
This and similar places. Second argument is for error reporting, so it should help find what’s wrong. I suggest something like:
```
(defn- resolve-datom [db e a v t]
  (when a (validate-attr a (list 'resolve-datom 'db e a v t)))
...
```
This way in error report you’ll see something like `Bad entity attribute 177 at (resolve-datom db 1 177 nil nil), expected keyword or string`

tonsky:
Decided to use `(and (instance? ISearch o) (instance? IIndexAccess o) (instance? IDB o))` for `db?` check

tonsky:
Can you move `make-obj-array` to `datascript.core` and rename to something shorter like `arr`? We’ll be using this a lot, probably. Also it’ll look nice to have `arr` and `array?` co-located.

tonsky:
`ICache` could be dropped. I already forgot what was my intention, it’s clearly not working in the current form

tonsky:
Please keep CLJS version of `mapa` in `query_v3`. It was made this way during optimization works. Clojure version could be slow for now. Also, in `query_v3`, `make-array` should be changed to `arr` of ours.

tonsky:
Can your serialization test be moved to `datascript.test.serialization`?

benfleis:
All looks good; serialization already moved in my branch, and I wondered about and debated putting a comment about `mapa`, but knew you'd see it and let me know if it was intentionally structured that way.

benfleis:
@tonsky
Was looking for an example of data_readers.clj, and came immediately upon this thread, saying that libraries shouldn't be defining it.

  https://groups.google.com/d/msg/clojure/B4_uGy1VhnA/pZdF3Eq3j5oJ

And yet, some do. Trivial for it to be there, or not, but this pushes me to think better of defining ds/data-readers as the map, and letting users do as they please. (CLJS has a register mechanism, and is fundamentally ... more compatible.)

tonsky:
- `:datoms` in DB are now double-wrapped in `[[]]`, they shouldn’t be:
```
      (is (= (pr-str db) "#datascript/DB {:schema nil, :datoms [[1 :foo \"bar\" 25]]}"))
```
- For clojure, we should check both `clojure.core/read-string` with default params and `clojure.edn/read-string` with `d/data-readers`
- in `-index-range`, please fix validate-attr the same way you did for `resolve-datom`
- you haven’t removed `filtered-db?` yet

Otherwise looks great. I’ll look into serialization/test issue

benfleis:
re `:datoms`: shouldn't it be exactly that, a vector of vectors? (or more generically, a seq of datoms, which are themselves seqs in this context?)

tonsky:
of course, terribly sorry, you’re right

benfleis:
using clojure.core read-string w/ data_readers.clj fails similarly to the commented out test below. i haven't yet solved this; if you've got an idea, let me know, otherwise this simply doesn't work atm :(

tonsky:
Oh, porting all these protocols is going to be very tough :(

query_v3:
```
(#?(:cljs array-reduce :clj reduce) ;; iterate over corresponding rel1 tuples
```
reduce and array-reduce have different argument order. Maybe create `dc/array-reduce` and use it here?

---
lookup_refs, parser, parser_query, parser_rules, and all the other places:
```
#?(:clj
   (import '[clojure.lang ExceptionInfo]))
```
Can you move it into ns declaration? I want to be as declarative as possible

---
```
 (defn- retract-components [db datoms]
-  (into #{} (comp
-              (filter #(component? db (.-a %)))
-              (map #(vector :db.fn/retractEntity (.-v %)))) datoms))
+  (->> datoms
+       (filter #(component? db (.-a %)))
+       (map #(vector :db.fn/retractEntity (.-v %)))
+       (into #{})))
```
Why changing transducers form to lazy seq one? We’re on 1.7 anyway

---
```
 (defn entity-map [db e]
   (when-let [entity (d/entity db e)]
     (->> (assoc (into {} entity) :db/id (:db/id entity))
-         (clojure.walk/postwalk #(if (de/entity? %)
-                                     {:db/id (:db/id %)}
-                                     %)))))
+         (clojure.walk/prewalk #(if (de/entity? %)
+                                  {:db/id (:db/id %)}
+                                  %)))))
```
Why change here?

---
```
;; use aget in cljs, get in clj
(def get-compat #?(:cljs aget :clj get))
```
This is still going through additional fn call. This place is very hot. Can you write it that way so CLJS version stays exactly what it was (putting conditionals directly in every call place, not abstracting to fn). 

---
I’m also not happy with this:
```
 (defn tuple-key-fn [getters]
   (if (== (count getters) 1)
     (first getters)
     (let [getters (to-array getters)]
       (fn [tuple]
-        (list* (.map getters #(% tuple)))))))
+        (list* (map #(% tuple) getters))))))
```
Maybe it needs a conditional too.

benfleis:
> `(#?(:cljs array-reduce :clj reduce) ;; iterate over corresponding rel1 tuples`
> reduce and array-reduce have different argument order. Maybe create dc/array-reduce and use it here?

I will try `areduce` instead.


> `#?(:clj (import '[clojure.lang ExceptionInfo]))`
> Can you move it into ns declaration? I want to be as declarative as possible

Easy peasy. I debated which was clearer in this context, since it was intended to echo the (def Exception js/Error) type things.


>  `(defn- retract-components [db datoms] ...)`
> Why changing transducers form to lazy seq one? We’re on 1.7 anyway

Transducer part was forgotten, and will be fixed. The order swapping that came with it was not accidental. Speaking to you offline about that. (After: reverting entirely. My misunderstanding.)


> `-         (clojure.walk/postwalk #(if (de/entity? %)`
> `+         (clojure.walk/prewalk #(if (de/entity? %)`
> Why change here?

If we use postwalk, then Entity needs to support assoc, which we don't want to do. See https://github.com/tonsky/datascript/pull/74#issuecomment-98659268


> `(def get-compat #?(:cljs aget :clj get))`
> This is still going through additional fn call. This place is very hot. Can you write it that way so CLJS version stays exactly what it was (putting conditionals directly in every call place, not abstracting to fn).
> ...
> `-        (list* (.map getters #(% tuple)))))))`
> `+        (list* (map #(% tuple) getters))))))`

Will do.

PR #91, opened 2015-06-20
Pull API map spec non-component reversed attributes should return collections

mattsenior:
Hello.

With the pull api, if you reverse lookup a cardinality-one, non-component attribute inside a map spec, it only yields a single result, but it should yield a collection instead.

Here’s a test snippet to try to clarify. On the current `master` the top two tests will pass, but the bottom two tests (with map specs) will fail.

```clojure
(def test-schema {:father { :db/valueType :db.type/ref }})

(def test-datoms
  (->>
    [[1 :name  "Petr"]
     [2 :name  "David"]
     [3 :name  "Thomas"]
     [6 :name  "Matthew"]
     [2 :father 1]
     [3 :father 1]
     [6 :father 3]]
   (map #(apply d/datom %))))

(def test-db (d/init-db test-datoms test-schema))

(deftest test-pull-reverse-attr-spec
  (testing "Reverse non-component references yield collections"

    ;; Currently these two will pass
    (is (= {:name "Thomas" :_father [{:db/id 6}]}
           (d/pull test-db '[:name :_father] 3)))

    (is (= {:name "Petr" :_father [{:db/id 2} {:db/id 3}]}
           (d/pull test-db '[:name :_father] 1)))

    ;; But these two will fail
    (is (= {:name "Thomas" :_father [{:name "Matthew"}]}
           (d/pull test-db '[:name {:_father [:name]}] 3)))

    (is (= {:name "Petr" :_father [{:name "David"} {:name "Thomas"}]}
           (d/pull test-db '[:name {:_father [:name]}] 1)))))
```

This PR passes the tests, and brings the behaviour in line with Datomic, e.g. in the standard Datomic `mbrainz-1968-1973` db, this will yield the country name and a collection of all artist names:

```clojure
(d/pull db '[:country/name {:artist/_country [:artist/name]}] :country/GB)
```

I’m new to DataScript so my fix might not be the right way to do this.

Cheers!

### DIFF
--- a/src/datascript/pull_api.cljc
+++ b/src/datascript/pull_api.cljc
@@ -113,9 +113,9 @@
                (cond->> datoms
                  limit (into [] (take limit))))]
     (if found
-      (let [multi?     (dc/multival? db attr)
-            ref?       (dc/ref? db attr)
+      (let [ref?       (dc/ref? db attr)
             component? (and ref? (dc/component? db attr))
+            multi?     (if forward? (dc/multival? db attr) (not component?))
             datom-val  (if forward? (fn [^Datom d] (.-v d)) (fn [^Datom d] (.-e d)))]
         (cond
           (contains? opts :subpattern)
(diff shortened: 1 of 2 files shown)


tonsky:
Great fix, thanks! Merged here https://github.com/tonsky/datascript/commit/a7eb43edf5f265a65d49a901f988fe56b8d3a7ab

PR #93, opened 2015-06-22
Add Node.js REPL.

dahjelle:
Since the browser REPLs are a bit clunky (and unnecessary) and LightTable's development has slowed down, I thought adding a Node.js REPL would be nice. (At least until `lein-cljsbuild` adds it.)

`lein noderepl` should launch it, provided you have Node installed somewhere.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -16,9 +16,11 @@
   
   :jvm-opts ["-Xmx2g" "-server"]
 
-  :aliases {"test-clj" ["run" "-m" "datascript.test/test-all"]}
-  
-  :cljsbuild { 
+  :aliases {
+    "test-clj" ["run" "-m" "datascript.test/test-all"]
+    "noderepl" ["run" "-m" "clojure.main" "repl.clj"]}
+
+  :cljsbuild {
     :builds [
       { :id "release"
         :source-paths ["src" "bench/src"]
--- a/repl.clj
+++ b/repl.clj
@@ -1,0 +1,13 @@
+(require 'cljs.repl)
+(require 'cljs.build.api)
+(require 'cljs.repl.node)
+
+(cljs.build.api/build "src"
+  {:main 'datascript
+   :output-to "target/datascript.js"
+   :output-dir "target"
+   :verbose true})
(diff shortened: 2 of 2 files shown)


dahjelle:
Rebased on top of latest changes and fixed the whitespace to clean up the diff a bit.

benfleis:
@dahjelle Hey, I don't speak for @tonsky obviously, but I might suggest making a 'dev' dir, and putting repl.clj in there, and updating the project.clj to include that as a src dir in the dev profile. (Locally I do this with a dev/user.clj, which contained my own hooks for node repl.) In fact, if you put that into user.clj, it will automatically be available from your repl, if that's the desired effect.
$0.02 -> done.

dahjelle:
Yeah, I'm quite new to ClojureScript, so wide open to suggestions of better ways of doing things. :-D I'm happy to update this pull request (if others would find it useful) to whatever the best way of doing it is.

Let me see if I understand your comment correctly:

1. In the `:dev` profile in `project.clj`, change the source paths to

```
:source-paths ["bench/src" "test" "dev"]
```
2. In the `dev` directory, put the `repl.clj`.

Would the `:aliases` section stay the same, except for the path? Or were you saying that is the bit that could be moved into `user.clj`? Is `user.clj` a file `leiningen` treats specially?

Sorry for all the questions—thanks again for any suggestions!

benfleis:
source-paths update is correct.

if user.clj exists in a source path, it is already loaded when you open a repl, since the repl starts in user namespace. thus, it's handy to have a user.clj contain things that you want to "just be available", without having to explicitly eval that file.

thus, if you want that code to be automatically available upon starting a normal repl, put it in user.clj instead of repl.clj.

make sense?

tonsky:
So I went with my own implementation b/c I needed to cover more cases (browser repl, test compilation). Check it out here: https://github.com/tonsky/datascript/commit/4b63a1a48366f702bc541eb07b86b4c2ae3054f4

dahjelle:
Awesome! Thanks!

PR #97, opened 2015-06-28
optional millisecond argument for squuid constructor

ThomasDeutsch:
optional millisecond argument for squuid constructor.


tonsky:
Decided to go with my own patch — yours has a bug, it was faster to fix it by myself: f2f05fa

PR #104, opened 2015-07-23
Fix sneaky parens caught by humane-test-output

aengelberg:
[humane-test-output](https://github.com/pjstadig/humane-test-output) plugin caught a common paren oversight when I tried to run the tests locally.


tonsky:
Thanks, a great catch!

PR #107, opened 2015-08-05
Bring explicitly supported value types in line with Datomic

sgrove:
It's valuable to be able to dump (at least a larger subset) Datomic schema into DataScript automatically. This change adds support for more types. The only dubious one is :db.type/double since that can't round-trip.

I don't believe there should be any performance impact, and this is a very handy change to have.

### DIFF
--- a/src/datascript/core.cljc
+++ b/src/datascript/core.cljc
@@ -514,6 +514,11 @@
                      :key k
                      :value v}))))
 
+(def supported-value-types
+  #{:db.type/keyword :db.type/string :db.type/boolean :db.type/long
+    :db.type/bigint :db.type/float :db.type/double :db.type/bigdec :db.type/ref
+    :db.type/instant :db.type/uuid :db.type/uri :db.type/bytes})
+
 (defn- validate-schema [schema]
   (doseq [[a kv] schema]
     (let [comp? (:db/isComponent kv false)]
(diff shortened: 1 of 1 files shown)


tonsky:
DS doesn’t know what to do with types anyways. Only use-case I can think of is to use schema just to preserve Datomic schema. In that case, it would be valuable to have all the types Datomic support. Can you add the rest? http://docs.datomic.com/schema.html#sec-1-1

sgrove:
Should there be any kind of dev-time warning about value types that DS doesn't understand? It might be nice to highlight that these are "advanced features" and can be dangerous if the dev doesn't understand the difference between Datomic/DS.

tonsky:
How do you see it? DS understands any type that is IComparable. I don’t want to limit that in any way.

sgrove:
This PR is probably good to go as is, warnings can go in a separate PR.

The warning would happen in dev at db instantiation time (or migration time, if DS ever supports that, as in #14 -  which would help a bit for Dato), and would just have a few cases where:

 1. ClojureScript/JavaScript doesn't have the type
 1. Round-tripping is impossible, even with e.g. transit-cljs  (e.g. doubles won't be serialized as such client-side)

This would just be to help people diagnose why the data they're getting from DS/transit isn't going into Datomic properly.

tonsky:
I spent some time thinking about it. I want this validation to be meaningful, do not lie, and be helpful. Round-tripping is just one, very specific use-case, I don’t want to design API around that single issue.

E.g. you can put vector as a value. DataScript allows that. It’s not a hack, it’s part of the contract. Why can’t you specify vector as a type in the schema?

On the other hand, there’s no long/int/floats in JS, just number which is 64 bits floating point. So it doesn’t actually makes sense to let you use them in the schema.

There will also be no validation once you’ve specified data type. This is counter-intuitive (why else would I specify data type if it isn’t validated later?)

Given all that, I think I’ll postpone this decision until real reason for types in the schema rises. Once we have a use-case for that, it’ll be clear how types in the schema should work, should be validation, should we allow for arbitrary types etc. Sorry it took so long.

PR #109, opened 2015-08-27
Add failing test for vector-based upsert failure

levand:


tonsky:
I fixed that, but wrote my own test, so closing this. Thanks for bringing my attention to the issue!

PR #114, opened 2015-09-15
Fix null pointer exception when `contains?` is called with an entity.

lynaghk:
Using `contains?` on an entity on the JVM caused a null pointer exception because `contains?` expects the `containsKey` method to return a boolean.
I've updated the implementation of Entity to do the right thing on the JVM just like what's done on the cljs implementation (using the not-found arity of key lookup).


tonsky:
Thanks! Great fix

tonsky:
Released as 0.12.2

PR #116, opened 2015-09-18
make conn? work with any atom-like entity in cljs

piranha:


piranha:
I couldn't find a protocol for Clojure's side, so I left `instance?` in place.

tonsky:
I think check should be `IDeref` for both Clojure and ClojureScript. We need conn to `deref` to DB, so we should check exactly for that.

piranha:
Ok, also I'm not sure why tests failed - looking at that.

tonsky:
Merged in 0.13.1

PR #127, opened 2015-11-05
Let old-value for db.fn/cas be nil

### LINKED ISSUE #126: :db.fn/cas should be able to use nil for old value
Transactions fail when performing db.fn/cas on an entity which does not have an old value for the compared attribute:
```clj
(def conn (d/create-conn))
(d/transact! conn [[:db/add 1 :weight 200]])
 ;; works as expected:
(d/transact! conn [[:db.fn/cas 1 :weight 200 300]])
;; crashes with message "Cannot store nil as a value at [:db.fn/cas 1 :foo nil :bar]":
(d/transact! conn [[:db.fn/cas 1 :foo nil :bar]]) 
```
Expected entity 1: `{:weight 300 :foo :bar}`

From Datomic docs (http://docs.datomic.com/transactions.html):
> You can use nil for the old value to specify that the new value should be asserted only if no value currently exists.

petterik:
Fixes #126 by not validating old-value (letting it be nil).

### DIFF
--- a/src/datascript/db.cljc
+++ b/src/datascript/db.cljc
@@ -976,7 +976,6 @@
                     _ (validate-attr a entity)
                     ov (if (ref? db a) (entid-strict db ov) ov)
                     nv (if (ref? db a) (entid-strict db nv) nv)
-                    _ (validate-val ov entity)
                     _ (validate-val nv entity)
                     datoms (-search db [e a])]
                 (if (multival? db a)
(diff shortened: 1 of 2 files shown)


tonsky:
Thanks!

tonsky:
Released as 0.13.3

petterik:
Sweet! Thanks!

PR #155, opened 2016-03-26
Avoid keywordize attributes in schema->clj(closes #153)

### LINKED ISSUE #153: :db.unique/identity doesn't merge with JS API
Marking a schema as containing a unique identity doesn't merge the entities with the JS API. I've created a reproducible test case at https://gist.github.com/terinjokes/a5003ab475a6fb5eddaa.

typeetfunc:
Keywordize attributes in schema->clj looks like a bug(because in other functions you avoid convert user string to keyword). See: https://github.com/tonsky/datascript/issues/153#issuecomment-200723583.

### DIFF
--- a/src/datascript/js.cljs
+++ b/src/datascript/js.cljs
@@ -14,7 +14,8 @@
 
 (defn- schema->clj [schema]
   (->> (js->clj schema)
-       (walk/postwalk keywordize)))
+       (reduce-kv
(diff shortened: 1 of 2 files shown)


tonsky:
Great, thanks! Nice contribution.

tonsky:
Just pushed 0.15.1

PR #158, opened 2016-04-21
Define db-with with comp

lvh:
This means you can also use db-with when you have tx-meta. The precondition shouldn't matter because with already has the same one.

### DIFF
--- a/src/datascript/core.cljc
+++ b/src/datascript/core.cljc
@@ -58,9 +58,8 @@
                                :tempids   {}
                                :tx-meta   tx-meta}) tx-data))))
 
-(defn ^:export db-with [db tx-data]
-  {:pre [(db/db? db)]}
-  (:db-after (with db tx-data)))
+(def ^:export db-with
+  (comp :db-after with))
 
 (defn ^:export datoms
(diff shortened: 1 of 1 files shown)


tonsky:
Do you have any use-case for that? If you use `db-with` with `tx-meta`, `tx-meta` will be ignored and thrown away.

lvh:
Sort-of. I have some code that wishes it could call both db-with and with with the same signature. I could see why you wouldn't want to take an unused argument though; but then perhaps you'd prefer a comment?

tonsky:
I think signature is enough, no? In what scenario would you see a comment
but wouldn’t see a signature?

lvh:
It's not so much that you would see one and not the other; just that if you don't know/forgot tx-meta is a transaction-level only thing (as opposed to tx-data, which affects the DB), that might be confusing since the two are obviously related fns; hence why I went for the comp first.

tonsky:
Can you give an example what kind of comment would it be? And where should
it be located?

lvh:
Having tried the comment, if this PR isn't something you think is reasonable then personally I think the comment is probably worse anyway, so I guess I'm closing this PR instead :)

tonsky:
Ok :)

PR #164, opened 2016-06-28
Filter assertions concerning transaction entities when determining ma…

### LINKED ISSUE #163: Entity and transaction id partitions not respected in `db-init`
At https://github.com/tonsky/datascript/blob/e1ca971762401444660924120cc6d9d0be58c7c0/src/datascript/db.cljc#L589 and https://github.com/tonsky/datascript/blob/e1ca971762401444660924120cc6d9d0be58c7c0/src/datascript/db.cljc#L595, the `max-eid` will be set into the transaction partition if their are assertions about transaction ids.  This isn't necessarily a bug, but I do think it's unwanted because it means that replaying transactions yields a database identical in all ways save the `max-eid` field.  That surprised at least me.

The fix is simple: filter the incoming ids to remove transaction ids.

@tonsky do you agree that this behaviour should be changed?

ncalexan:
…x-eid. (closes #163)

This patch ensures that max-eid is not set into the transaction
partition if there are assertions about transaction ids in the initial
collection of datoms provided to db-init.

This means that replaying transactions yields a database identical in
all ways, including the max-eid field.

### DIFF
--- a/src/datascript/db.cljc
+++ b/src/datascript/db.cljc
@@ -586,13 +586,17 @@
                                          #js [] datoms)
                                  (.sort cmp-datoms-avet-quick))
                  avet    (btset/-btset-from-sorted-arr avet-datoms cmp-datoms-avet)
-                 max-eid (:e (first (-rseq eavt)))]
+                 max-eid     (:e (first (-rseq (btset/slice eavt
+                                                           (Datom. nil nil nil nil nil)
+                                                           (Datom. (dec tx0) nil nil nil nil)))))]
                 :clj
                 [eavt        (apply btset/btset-by cmp-datoms-eavt datoms)
                  aevt        (apply btset/btset-by cmp-datoms-aevt datoms)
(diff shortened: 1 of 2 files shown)


tonsky:
Awesome! Thanks

tonsky:
Just pushed 0.15.1

ncalexan:
@tonsky you're welcome!  We love DataScript and have learned a lot from it.  We (a team at Mozilla) are starting a small project to build a similar Datomic-API-alike over at https://github.com/mozilla/datomish.  Early days yet but we'd love to have your thoughts.

tonsky:
No access

ncalexan:
@tonsky sorry about that -- didn't realize we hadn't opened it up yet.  I've added you as a collaborator.

tonsky:
Nice! I already smell CREATE TABLE EAVT; :))

Do you plan to reuse any parts of DataScript code? I think query engine
would be nice to have, but I’m not sure how easy it’d be to reuse it. Do
you plan queries at all?

Issue #172, opened 2016-08-11
Transaction data with ref attributes is order dependent (differs from Datomic)


kennyjwilli:
Set up minimal example db:
```clojure
(def uuid1 #uuid"57abca79-11ff-46cc-884d-3fb72d51d1d4")
(def uuid2 #uuid"57abca79-fd2a-49a6-8437-d9ce80a61c49")
(def db (d/db-with
          (d/empty-db {:entity/id {:db/unique :db.unique/identity}
                       :child     {:db/valueType :db.type/ref}})
          [[:db/add -1 :entity/id uuid1]
           [:db/add -2 :entity/id uuid2]
           [:db/add -2 :child -1]]))
```
This works, as expected: 
```clojure
(d/db-with db [[:db/retract 1 :entity/id uuid1]
               [:db/retract 2 :child 1]])
```

The fails with this exception `"Nothing found for entity id [:entity/id #uuid \"57abca79-11ff-46cc-884d-3fb72d51d1d4\"]"`
```clojure
(d/db-with db [[:db/retract 1 :entity/id uuid1]
               [:db/retract 2 :child [:entity/id uuid1]]])
```

In the failing case, the first datom removes the `:entity/id` attribute and the second datom refers to the now removed `:entity/id`. Because Datascript materializes each intermediate database, this set of tx-data will not work. Datomic does not have this issue. 

In [this issue](https://github.com/tonsky/datascript/issues/76#issuecomment-98647015) you mention that this is a design decision. 
> This won’t work in current design, because in DS every transaction clause is executed one after another. Each intermediate DB state is materialized in between.

I would argue that this is a bad design decision. The assumption should be that each transaction datom is transacted against the current value of the database, not some internal, intermediate value. Therefore, the transaction should be defined as a set of data that occurs all at one time. This is especially concerning because you would expect the same result when you replace the entity ids with lookup refs in the transaction data.

The workaround, for now, is to sort the tx data such that any retracted `:v` lookup refs occur before retracted `:e` lookup refs.

tonsky:
Yes, downside of DataScript approach is that you have to think about order of clauses in a transaction. That probably hurts most when doing bunch of retractions.

[...]

PR #173, opened 2016-08-12
Use the initial DB for resolving lookup refs when retracting datoms

sparkofreason:
Use the initial DB for resolving lookup refs when retracting datoms in a transaction. Fixes tonsky/datascript#172


sparkofreason:
Not sure how to avoid the massive whitespace changes caused when wrapping an s-expression. The relevant changes are at lines 998 and 1105.

tonsky:
Closing since I don’t plan to fix #172

PR #178, opened 2016-09-14
Var resolution for query fns/preds when in JVM

levand:
This commit adds a reader conditional to enable the use of
namespace-qualified symbols as query functions or predicates, when in
JVM Clojure. This allows queries using functions or predicates to be
reused between Datomic and DataScript when running on the JVM.

The semantics of a DataScript query are otherwise unchanged; global var
resolution is only attempted as a final step before failure.

The semantics when operating in a ClojureScript context are entirely
unchanged, since global symbol->var resolution is not supported.

### DIFF
--- a/src/datascript/query.cljc
+++ b/src/datascript/query.cljc
@@ -430,10 +430,16 @@
                              args)]
       (apply f resolved-args))))
 
+(defn- resolve-sym [sym]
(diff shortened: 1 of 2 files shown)


tonsky:
Thanks, good stuff!

PR #179, opened 2016-09-17
Added js->clj usage to datoms/seek-datoms/q

serebrianyi:
Added js->clj to fix the js interface when using lookup refs

### DIFF
--- a/src/datascript/js.cljs
+++ b/src/datascript/js.cljs
@@ -69,7 +69,8 @@
 
 (defn ^:export q [query & sources]
   (let [query   (cljs.reader/read-string query)
-        results (apply d/q query sources)]
+        js-sources (map #(js->clj %) (rest sources))
+        results (apply d/q query (conj js-sources (first sources)))]
     (clj->js results)))
 
 (defn ^:export pull [db pattern eid]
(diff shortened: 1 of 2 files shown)


tonsky:
I think you’re using JS API wrong.

`datoms`/`seek-datoms` accept variable arguments, not an array. Change `d.datoms(db, ":eavt", ["name", "Ivan"])` to `d.datoms(db, ":eavt", "name", "Ivan")`. There’re two tests showing it https://github.com/tonsky/datascript/blob/ba1246704c1dfe48a5f0abc330a5098b5f79edce/test/js/tests.js#L429-L434

Queries over native JS arrays work well too. Check these tests out https://github.com/tonsky/datascript/blob/ba1246704c1dfe48a5f0abc330a5098b5f79edce/test/js/tests.js#L347-L379

I’m not sure what was your intention doing this:
```
assert_eq("Ivan", d.q('[:find ?c . :in $ ?e :where [?e "name" ?c]]', db, ["name", "Ivan"]));
```
but if you can describe to me what were you trying to do I might be able to help you write the query for that.

serebrianyi:
Sorry, I might really misunderstand something, but it doesn't seem like ```d.datoms/d.seek-datoms/d.q``` accept lookup refs right now, but ```d.entity``` does: ```assert_eq("Ivan", d.entity(db, ["name", "Ivan"]).get("name"));```
```d.datoms(db, ":eavt", "name", "Ivan")``` actually fails with ```"Expected number or lookup ref for entity id, got \"name\""```

The query I used is just a random one to test lookup refs inside of queries.

And thanks a lot for datascript, it's a really wonderful thing!

tonsky:
You’re right, lookup refs didn’t work from JS API in index APIs an queries. I didn’t realise you were trying to use them because for them to work you need to mark your attributes as `:db/unique` in schema. Just fixed that & added correct test cases. See https://github.com/tonsky/datascript/commit/4b7e2c447da7253be5eb846b231ab831542aba7e

serebrianyi:
Perfect, thank you!

PR #180, opened 2016-09-24
possibly a typo

alexei-matveev:
Spotted (assoc production [:tuples] ...) at two locations which are likely typos
in dead branches.

The other staff did not help much.

### DIFF
--- a/src/datascript/query.cljc
+++ b/src/datascript/query.cljc
@@ -420,15 +420,30 @@
     [(update-in context [:rels] #(remove (set rels) %)) production]))
 
 (defn -call-fn [context rel f args]
-  (fn [tuple]
-    ;; TODO raise if not all args are bound
-    (let [resolved-args (map #(if (symbol? %)
-                                (or
-                                 (get (:sources context) %)
-                                 (#?(:cljs aget :clj get) tuple (get (:attrs rel) %)))
-                                %)
-                             args)]
-      (apply f resolved-args))))
+  ;; The closure we  return from here appears to be  called a lot, try
+  ;; resolving as much as we can beforehand.  Any argument that is not
+  ;; a symbol resolves  to itself.  A symbol could be  resolved from a
+  ;; context  map, and  only if  it  does not  (i.e. is  falsy in  the
+  ;; original  version) the  resolution  has to  be  postponed to  the
+  ;; execution time of the closure.
+  (let [context-sources (:sources context)
+        rel-attrs (:attrs rel)
+        ;; Try resolving arguments from context:
+        args (mapv (fn [a]
+                     (if-not (symbol? a)
+                       [:resolved a]
+                       (if-let [v (get context-sources a)]
+                         [:resolved v]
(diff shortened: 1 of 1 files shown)


tonsky:
Thanks for raising the issue.

Yes, that code could be called in some circumstances (see https://github.com/tonsky/datascript/blob/7dada20fa98df268fdb05083638781ae7d63b3b3/test/datascript/test/query_fns.cljc#L262-L268). It’s kind of a hack to call predicates that way, and will be removed at some point in the future, but so far it’s possible and better to have it fixed, so thank you!

As of speed improvement, your changes didn’t do much of a difference, but I came up with my own version that gives 25-30% boost for predicate/fn calls in queries. See https://github.com/tonsky/datascript/blob/7dada20fa98df268fdb05083638781ae7d63b3b3/src/datascript/query.cljc#L422-L441

alexei-matveev:
Hi! Does the public API rely on tuple-idx being strings ("e", "a", ..)?
Because when with "nth" instead of "get" here: (#?(:cljs aget :clj nth) tuple tuple-idx) I get another 10 % speed up in the code below. (At the expense of breaking tests, of course).

```
;; This is significantly slower than SQLite code below:
(defn- bench-O2 [n]
  (let ;; rules:
      [r (quote [[(follows ?a ?b)
                  [?a ?b]]
                 [(follows ?a ?b)
                  [?a ?x]
                  (follows ?x ?b)]])
       ;; query:
       q (quote [:find ?a ?b
                 :in $ %
                 :where (follows ?a ?b)])
       ;; data:
       $ (mapv vector (range n) (map inc (range n)))]
    (d/q q $ r)))

;; (time (count (bench-O2 200)))
;; "Elapsed time: 7549.735629 msecs"
;; "Elapsed time: 6100.434851 msecs" (after addressing PR-180)
;; "Elapsed time: 5434.237636 msecs" (after get -> nth)
;; 20100

;; drop table if exists db;
;; create table db (x integer, y integer);

;; insert into db
;; with recursive
;;      cnt (x, y) as
;;      (values(0, 1)
;;         union all
;;       select
;;         x + 1,
;;         x + 2
;;       from cnt
;;       where x + 1 < 200)
;; select x, y from cnt;

;; select count(*) from (
;; with recursive
;;      follows (a, b) as
;;      (select x as a, y as b from db
;;      union all
;;      select f.a as a, d.y as b
;;      from follows as f
;;      join db as d
;;      on f.b = d.x)
;; select a, b from follows
;; )
;; ;

;; sqlite> .read recursive.sql
;; Run Time: real 0.000 user 0.000000 sys 0.000000
;; Run Time: real 0.000 user 0.000000 sys 0.000000
;; Run Time: real 0.001 user 0.004000 sys 0.000000
;; count (*)
;; 20100
;; Run Time: real 0.047 user 0.044000 sys 0.00000
```

tonsky:
I believe it does, yes. Queries and recursive rules especially are some of the least optimized parts of DataScript, plenty of other opportunities there to speed that stuff up

PR #185, opened 2016-10-30
Added installed transaction functions using :db/ident and :db/fn semantics

refset:
The existing approach uses function references with :db.fn/call which meant I couldn't trivially serialise  (using transit/write) tx-data before expansion into datoms, so I created this alternative which should also enable some interesting new patterns for function modularity. I appreciate it's probably not "native" enough to be accepted as-is, but what do you think? Do you see a better way?

### DIFF
--- a/src/datascript/db.cljc
+++ b/src/datascript/db.cljc
@@ -1109,8 +1109,8 @@
               (and (ref? db a) (tx-id? v))
                 (recur (allocate-eid report v (current-tx report)) (cons [op e a (current-tx report)] entities))
 
-              (tempid? e)
-                (if (not= op :db/add)
+              (and (tempid? e) (some #(= op %) [:db/add :db.fn/call :db.fn/cas :db/cas :db/retract :db.fn/retractAttribute :db.fn/retractEntity]))
+                (if (some #(= op %) [:db.fn/call :db.fn/cas :db/cas :db/retract :db.fn/retractAttribute :db.fn/retractEntity])
                   (raise "Tempids are resolved for :db/add only"
                          { :error :transact/syntax
                            :op    entity })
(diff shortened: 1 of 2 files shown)


tonsky:
Interesting idea! I think, this should be done a little more in line with how Datomic works. E.g. using `:db/ident` for function name (keyword), `:db/fn` for actual function value (`IFn`) and `[<ident> & <args>]` syntax for calling it in a transaction. Do you mind making these changes?

refset:
Okay I think I've just about managed it! How does that look?

refset:
Sorry, I inadvertently (automatically!) closed this when trying to sort out the branches in my fork. I'd be happy to try to rebase these changes sometime.

src/datascript/db.cljc
```diff
@@ -1109,8 +1109,8 @@
               (and (ref? db a) (tx-id? v))
                 (recur (allocate-eid report v (current-tx report)) (cons [op e a (current-tx report)] entities))
 
-              (tempid? e)
-                (if (not= op :db/add)
+              (and (tempid? e) (some #(= op %) [:db/add :db.fn/call :db.fn/cas :db/cas :db/retract :db.fn/retractAttribute :db.fn/retractEntity]))
```
tonsky: Hey @refset, what those two lines are about? Why are they necessary?

refset: Hey! Yeah it's a pretty unfortunate solution but I haven't been able to think of a clearer way to solve it. Essentially the problem is that with this :db/ident semantic the let binding `(let [[op e a v] entity]` means op is the :db/ident but e could be anything (including a string), because it is simply the first argument for the :db/fn function. If the first argument happens to be a string or a negative number then `(tempid? e)` returns true, when it needs to be false if op is a :db/ident. These two lines essentially mean `(and (tempid? e) (not (custom-op? op))`. Note that :db/add doesn't appear in the vec on the second line. Perhaps this is worth pulling into a new function, somehow...

PR #186, opened 2016-11-02
Built in vector, hash-map

claj:
I'm missing `vector` and `hash-map` as built in functions. I find them useful to construct new data structures that could be transacted, usually in combination with rules. The tests are meant to give a hint of how this could be useful.


tonsky:
Thanks! I’m adding `array-map`, `list` and `set` for completeness too https://github.com/tonsky/datascript/commit/b864f53644d0e77ec3134fba24afb6ca1c17103e

PR #192, opened 2016-12-09
Speedup rschema

rauhs:
Speedup rschema which helps faster load times for use in clojurescript with large schemas.

### DIFF
--- a/src/datascript/db.cljc
+++ b/src/datascript/db.cljc
@@ -513,27 +513,31 @@
 ;; ----------------------------------------------------------------------------
 
 (defn attr->properties [k v]
-  (cond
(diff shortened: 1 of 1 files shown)


rauhs:
I have a schema with about 100 attributes (and more to come) with a largish SPA. On my fast desktop the db init can take around 100ms but I also target mobile and then the init becomes 2-3x slower. One slow component was rschema during profiling, hence the PR.
The code is a little less readable due to not using for.

rauhs:
I've since switched to precomputing the rschema (in a macro) to further speedup startup. So this is now obsolete.

tonsky:
I’ve merged it anyways. Seems impactful

PR #205, opened 2016-12-28
Optimization: Add type hints for Clojure in `datascript/arrays`

wbrown:
The compiler appears to have difficulty discovering reflection when macros are used.

This addresses reflections in:
* `datascript.arrays/acopy`
* `datascript.arrays/aconcat`

Some optimizations were also rolled into `datascript.arrays/aconcat` as well.


tonsky:
How exactly did you trigger those warnings? I tried to reproduce myself but didn’t see any

tonsky:
I also didn’t see any noticeable perf increase

PR #216, opened 2017-04-11
fix externs.js and the exported symbols

thheller:
The `datascript/externs.js` is written incorrectly.

Including the externs via `deps.cljs` means that CLJS projects can never rename them as well which  isn't useful. CLJS projects that want to use this would be better off without the externs/exports.

You should consider only using the externs/exports when building the JS lib.

### DIFF
--- a/src/datascript/db.cljc
+++ b/src/datascript/db.cljc
@@ -183,6 +183,8 @@
         (assoc [d k v] (assoc-datom d k v))
         ]))
 
+#?(:cljs (goog/exportSymbol "datascript.db.Datom" Datom))
+
 (defn ^Datom datom
   ([e a v]          (Datom. e a v tx0 true))
   ([e a v tx]       (Datom. e a v tx true))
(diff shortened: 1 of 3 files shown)


tonsky:
Sure, but DataScript implementation depends on datom fields not being renamed. I have to include those. As for entity, what I was doing is aimed for JS only, that’s right.

zarkone:
Not sure if that related but I also had problems with `datascript`, compiling with `shadow-cljs`.  

For example, this code works without optimizations, but doesn't work with `:advanced` enabled, i.e. in `:realease` mode:
```
(d/q '[:find [?atom ...] :in $ ?url
                         :where [?e :url ?url]
                                [?e :atom ?atom]]
                   @db url)
```
I've solved it without patching anything by adding contents of  `datascript/externs.js` to my `manual-externs.js` file I already had for externs: 

Related part of `shadow-cljs.edn`:
```
:release {:compiler-options {:externs ["manual-externs.js"]
                             :optimizations :advanced}}
```

tonsky:
@zarkone are you sure it’s datascript problem? maybe your build just doesn’t know how to lookup externs served with libraries?

PR #228, opened 2017-06-13
Fixed deref of nested structures

serebrianyi:
Currently the JS version doesn't support dereferencing of existing attributes in a map form, e.g. on schema 
```
{"aka": {":db/valueType": ":db.type/ref", ":db/cardinality": ":db.cardinality/one"}}
``` 
transact with 
```
{ ":db/id": 1, "name": "Ivan" }, { ":db/id": 2, "name": "X" }, { ":db/id": 1, "aka": {":db/id": 2 }}
``` 
will result not in 2 entities, but 3 (one being created automatically by `db/transact-tx-data` since there is a check for :db/id as a keyword, but not as a string).
IMHO the reason is that only the :db/id on the main entity gets replaced by `js/entity->clj` and it doesn't happen recursively for the nested entities.

### DIFF
--- a/src/datascript/js.cljs
+++ b/src/datascript/js.cljs
@@ -19,11 +19,14 @@
 
 (defn- ^:declared entities->clj [entities])
 
+(defn- keywordize-db-id [e]
+  (let [f (fn [[k v]] (if (= ":db/id" k) [:db/id v] [k v]))]
+    (walk/postwalk (fn [x] (if (map? x) (into {} (map f x)) x)) e)))
+
 (defn- entity->clj [e]
-  (cond (map? e)
-    (-> e
-      (dissoc ":db/id")
-      (assoc  :db/id (e ":db/id")))
+  (cond
+    (map? e)
+      (keywordize-db-id e)
     (= (first e) ":db.fn/call")
       (let [[_ f & args] e]
         (concat [:db.fn/call (fn [& args] (entities->clj (apply f args)))] args))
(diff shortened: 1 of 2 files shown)


tonsky:
Wait, but transacting `{ ":db/id": 1, "aka": {":db/id": 2 }}` shouldn’t work in CLJ/CLJS either. You need to use entity id, not map: `{ ":db/id": 1, "aka": 2}`

serebrianyi:
Are you sure that it wasn't supposed to work? 
Because if I run the CLJ/CLJS version, it works just fine I think
```
(let [db  (-> (d/empty-db {:aka { :db/cardinality :db.cardinality/one 
                                  :db/isComponent true
                                  :db/valueType :db.type/ref}})
                (d/db-with [{:db/id 1 :name "Ivan"}])
                (d/db-with [{:db/id 2 :name "Petr"}])
                (d/db-with [{:db/id 1 :aka {:db/id 2}}]))]
   (d/pull db '[*] 1))
```
results in  `{:db/id 1, :aka {:db/id 2, :name "Petr"}, :name "Ivan"}`, which is also consistent with Datomic as far I understand.
What I'm actually trying to do is to transact to Datascript the result of Datomic pull - it works in Datomic, but was not working in the JS version of Datascript.
Or am I missing something completely here?

tonsky:
Well I need some time to load up all the details on how this is supposed to work :) I’ll check with Datomic and if it’s a legitimate behaviour I’ll see that it’s supported in all platforms including JS

serebrianyi:
Thank you!

tonsky:
Oops. I didn’t notice you were proposing a PR. I thought it’s an issue and fixed it myself :) Gave you credit in CHANGELOG though

Issue #219, opened 2017-04-19
0.16.0 broke :direct-linking for JVM version


tonsky:
With [datascript "0.16.1"] call to `Datom/nth` compiled with `{:direct-linking true}` will throw

```
Exception in thread "main" java.lang.NoSuchMethodError: datascript.db$nth_datom.invokeStatic(Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;
	at datascript.db.Datom.nth(db.cljc:128)
	at clojure.lang.RT.nth(RT.java:868)
	at reproduce.core$test_case.invokeStatic(core.cljc:7)
	at reproduce.core$_main.invokeStatic(core.cljc:9)
	at reproduce.core$_main.doInvoke(core.cljc:9)
	at clojure.lang.RestFn.invoke(RestFn.java:397)
	at clojure.lang.AFn.applyToHelper(AFn.java:152)
	at clojure.lang.RestFn.applyTo(RestFn.java:132)
	at reproduce.core.main(Unknown Source)
```

tonsky:
repo https://github.com/shparun/datascript_0.16.1_declared_bug
possible cause https://github.com/tonsky/datascript/commit/19bdbd1beb5d027f5cc99e1c006002f2ac6bbd3a

reported by: @shparun

petterik:
@shparun showed that if a type hint is added to the `^:declared` `defn`, then it won't throw NoSuchMethodException.
I added a case showing that the `^:declared` `defn` only needs to match primitive type hints:
https://github.com/petterik/datascript_0.16.1_declared_bug/commit/8143a6248f8a8159c680ea21a853ae3a551aa423

I went through all the `^:declared` `defn`s and type hinted the ones that had primitive "real" `defn`s to create #230

petterik:
Getting entities by lookup ref also fails when compiling with direct-linking true.
Test throws assertion error:
```clj
(defn get-unique-by-value-test-case []
  (let [conn (d/create-conn {:uni {:db/unique :db.unique/identity}})]
    (d/transact conn [{:uni :foo}])
    (assert (some? (:db/id (d/entity (d/db conn) [:uni :foo]))))))
```
Does not fail with direct-linking and `datascript 0.15.5`.
Does not fail with `datascript 0.16.1` and direct-linking false.
Fails with `datascript 0.16.1` and direct-linking true.

Verified that my pull-request #230 does not fix this issue.

Commit for the new test case:
[...]

PR #230, opened 2017-06-17
GH-219 Type hints all primitives for the :declared vars

petterik:
Adds primitive type hints to `^:declared` defns.

#219


petterik:
Does not entirely fix #219. Closing.

PR #232, opened 2017-06-20
Use goog.async.nextTick in place of setTimeout for transact-async

colindresj:
`goog.async.nextTick` is faster than `setTimeout`, so this updates `transact-async` to use it instead.

See https://groups.google.com/forum/#!topic/closure-library-discuss/hNs3W_bgnso

### DIFF
--- a/src/datascript/core.cljc
+++ b/src/datascript/core.cljc
@@ -5,7 +5,8 @@
     [datascript.pull-api :as dp]
     [datascript.query :as dq]
     [datascript.impl.entity :as de]
-    [datascript.btset :as btset])
+    [datascript.btset :as btset]
+    #?(:cljs goog.async.nextTick))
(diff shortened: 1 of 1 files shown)


tonsky:
Well, `transact-async` was only added for Datomic API compatibility. I don’t think there’s any value for it besides that. If you feel that you need to squeeze more perf out of it, just wrap normal transact into nextTick. Unless you have a strong case for nextTick to being there, of course

colindresj:
The scenario I'm thinking of is syncing with a datomic db. Or, really any large batch of transactions that need to be processed. Saving those few ms on every loop could help if you're doing other UI rendering not based on the data in datascript.

Wrapping `transact` would definitely work, but just updating `transact-async` felt more natural to me, and it's relatively low-cost code-wise.

Issue #193, opened 2016-12-09
Allow skipping schema validation


rauhs:
I'm currently optimizing the startup time of datascript and one more slow part (besided my PR for `rschema`) is the --in production unnecessary-- schema valdation. Could this be wrapped in a `(when ^boolean goog.DEBUG)` call and removed if debugging is off? I.e. the app is in production.

Issue #174, opened 2016-08-19
Editable schema


metasoarous:
I have an implementation of this in Datsync ([here](https://github.com/metasoarous/datsync/blob/master/src/dat/sync/client.cljc#L437)), and I'd really rather have it just be a function in `datascript.core` if you're amenable. More or less, it creates a new database from the old datoms and the updated schema.

There's a question about what the should be sent to listeners when this happens. It could perhaps be a more or less empty transaction report with keys `:schema-changes`, `:new-schema` and `:old-schema`, perhaps.

Happy to clean this up and PR if you like.

tonsky:
Replacing schema is trivial to do in user code. What I don't want to do is to provide schema altering fn that will do that and ignore all the issues that arise from the fact that data stored in DS might not suit new schema (unique values, ref type, arities). If I ever provide such fn, it should deal with all the issues (~the same way Datomic deals with them, checking constraints, raising errors, not allowing certain migrations, etc). I'm not against changing schema, but for now only user of DS library is responsible for all the inconsistencies. That's why I think such a fn should not be included in DS yet (but all the tools to build it yourself are there already)

metasoarous:
Fair enough. In my case (datsync), the only schema changes that come through are ones that had already passed through Datomic, so they'd more or less been "vetted". But you're right that there shouldn't be surprises in something that's a core part of datascript, and it would be easy for someone not thinking about the consequences to goof. Here are my thoughts on these issues:

* ref vs other types: Datomic actually doesn't let you change `:db/valueType`. So we could scan the datoms and make sure no attributes are being set as reference attributes when they've already been used for other data (and vice versa).
* uniqueness: Changing from unique to non-unique is pretty straight forward with Datomic, but the other direction is only allowed in certain cases (values are already unique and have an index, or there are no values). We could mimick that or simply disallow depending on how much work we're willing to put into it.
* cardinality: cardinality one to many is straight forward, but many to one requires deciding which value to keep. I think this may actually be straightforward; Under the new schema I think passing in the old datoms would deduplicate all but one of the values. If I'm wrong about that we could either disallow once there are multiple values, or filter all but the most recently asserted value.

Here are the relevant Datomic docs for reference: http://docs.datomic.com/schema.html#Schema-Alteration.

[...]

Issue #150, opened 2016-03-21
Connection (or ref? transactor?) protocol


metasoarous:
Such a protocol would make it possible to build custom connection/ref objects with special lifecycle functionality and capabilities. Right now the connections is just an atom, but it would be nice (for example) to be able to implement it as a Reagent `r/atom`. This could be helpful for some things we're thinking about for Posh (reactive filters and the like).

typeetfunc:
@metasoarous What you think about stream(or property) interface(RxJS) for connection? [See rxjs datascript integration example](https://github.com/typeetfunc/rx-datascript/blob/master/README.md#example) (WIP version)
There are many good reactive library(Rx, Most, Kefir etc). Integration Datascript with any of these very simple.

metasoarous:
Yeah; that's exactly the kind of thing I think this would make simpler to implement. @mpdairy has already worked up quite a bit of really great functionality in https://github.com/mpdairy/posh for connecting DataScript to Reagent. So that's not the problem really... It's just that in thinking about additional features for posh (and other libs we're working on, like https://github.com/metasoarous/datsync), we've realized having a more extensible notion of what it means to be a connection would be helpful for implementing certain features.

tonsky:
You can just extend IDeref (cljs) or implement IDeref (clj) and make sure it has DB inside:

```clj
(defn ^:export conn? [conn]
  (and #?(:clj  (instance? clojure.lang.IDeref conn)
          :cljs (satisfies? cljs.core/IDeref conn))
    (db/db? @conn)))
```
Would that work?

metasoarous:
Just realized I never responded back on this thread...

The one thing that's not dealt with seamlessly by simply extending/implementing IDeref is the transaction listener functionaliy. Since libraries like posh depend on listeners for doing their thing, it would be nice if a Conn (or whatever) protocol reflected some abstraction around this listener functionality. I'm not sure what that would be yet, and this also isn't super high priority for us right now, so no pressure on our end. For now just wanting to push the conversation forward in case others decide to take up the charge.

tonsky:
You get me confused. You want to keep listeners functionality but re-implement Conn as a protocol? What's the idea behind this? How will it be different from DataScript's Conn?

metasoarous:
> How will it be different from DataScript's Conn?

[...]

PR #234, opened 2017-08-03
added tx-middleware capability and schema-middleware

bamarco:
There are some issues we are solving for datsync to make the datomic/datascript transition smoother and I wanted to get your thoughts on the possiblity of adding tx-middleware to datascript.

This would address issues #193 and #174 
We will be using something along these lines in datsync, but I think it fits better in datascript itself.
- we still need to put the actual validation in as describe in #174 but the stub for it is hooked in
- In init-db I added an options map to turn off validation when it is undesirable which would allow people affected by #193 to make their own create-conn function in user space that skips validation
- tx-middleware will allow you to use `schema-middleware` to check txs for schema changes and put them into the `schema` map, but this would also open up capabilities like adding validation middleware for all the datomic :db/valueType for people who want type safety rather than speed and other things of that nature
- right now i put the `schema-middleware` in `datascript.db` but it could be put in its own ns if you think that makes more sense
- brings us closer to full ident support without performance hits on people who don't need that

Usage would look like
```clj
(defn test-schema-middleware []
  (let [conn (ds/create-conn)]
    (ds/transact!
      conn
      [{:db/ident :test
        :db/cardinality :db.cardinality/many}]
      {:datascript.db/tx-middleware datascript.db/schema-middleware})
    (ds/transact!
      conn
      [{:db/id -1
        :test :a}
       {:db/id -1
        :test :b}])
    @conn))
```

### DIFF
--- a/src/datascript/db.cljc
+++ b/src/datascript/db.cljc
@@ -26,7 +26,7 @@
           data (last fragments)]
       `(throw (ex-info (str ~@(map (fn [m#] (if (string? m#) m# (list 'pr-str m#))) msgs)) ~data)))))
 
-(defn #?@(:clj  [^Boolean seqable?]  
+(defn #?@(:clj  [^Boolean seqable?]
           :cljs [^boolean seqable?])
   [x]
   (and (not (string? x))
@@ -143,7 +143,7 @@
         IIndexed
         (-nth [this i] (nth-datom this i))
         (-nth [this i not-found] (nth-datom this i not-found))
-        
+
         IAssociative
         (-assoc [d k v] (assoc-datom d k v))
 
@@ -168,7 +168,7 @@
         (empty [d] (throw (UnsupportedOperationException. "empty is not supported on Datom")))
         (count [d] 5)
         (cons [d [k v]] (assoc-datom d k v))
-        
+
         clojure.lang.Indexed
         (nth [this i]           (nth-datom this i))
         (nth [this i not-found] (nth-datom this i not-found))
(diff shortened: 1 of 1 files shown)


metasoarous:
FWIW @tonsky, this was the original inspiration behind #150. At the time, Datsync and Posh were both needing to override various behavior of vanilla DataScript, and thought protocols might simplify things. Upon further reflection though, much of the behavior we wanted to override could be expressed via transaction middleware. I think if we get the shape of the tx-middleware idea right, it will be a much more composable way of getting much of the desired flexibility.

@bamarco Would you please describe the signature of the `tx-middleware` function to clarify how it handles database, tx-data and tx-report?

@tonsky Please let us know if you have any questions or concerns.

bamarco:
As currently written, `tx-middleware` is a function which takes a `transaction-fn` and returns a `transaction-fn`. The `transaction-fn` has the signature of `datascript.db/transact-tx-data` which takes in an `initial-tx-report` and some `txs`. 
```clj
(defn schema-middleware [transact]
  (fn [report txs]
    (let [{:as report :keys [db-after tx-data]} (transact report txs)
          db-after' (transduce
                    (filter schema-datom?)
                    conj-schema-datom
                    db-after
                    tx-data)]
      (if (= (:schema db-after) (:schema db-after'))
        report
        (assoc report
          :db-after (replace-schema db-after (:schema db-after')))))))
```

The above is an example. I have a suspicion that I actually need to call transact once more in that function for it to flow properly to the rest of the `tx-middleware`'s that could be `comp` with `schema-middleware`. `transact` is the `transaction-fn` being transformed and it would be called again like `(transact report [])` if needed. There also might be a way to make it act more plain transducers, but i haven't thought through that possibility.

PR #245, opened 2018-01-06
added the ability to use keyword refs directly as entity ids

bamarco:
This is a more modest update than tx-middleware. It is an important piece for syncing schema between different databases. This would be too difficult to do outside of datascript.
  
All it is doing is translating `:keyword` used in eid positions to `[:db/ident :keyword]` style lookup-refs before resolving. It will error if you don't have `:db/ident` in your schema.


fdserr:
I don't know about "syncing schemas", but that'd greatly improve porting queries and transactions to/from Datomic. Any new on this?

Issue #253, opened 2018-02-21
rseek-datoms


refset:
A desire for "`rseek-datoms`" has been previously discussed in the context of Datomic: https://groups.google.com/forum/#!topic/datomic/YkMn1z0o1Ao

Whilst I do not fully comprehend Stu's comment about time semantics, I can't imagine how that objection can apply to DataScript.

I figured out a temporary workaround using `btset/slice` directly:
```clojure
(reverse
  (btset/slice
    (get db :eavt)
    (d/datom. nil nil nil nil nil)
    (d/datom. eid a v nil nil)))
```
However, I would prefer to create a PR for `rseek-datoms` if you agree it's a worthwhile addition to the API...?

tonsky:
I honestly never understood value of `seek-datoms`. It’s basically “iterate from X till the very end of index”, which will include all sort of garbage at some point. Maybe you can give me examples?

The value of `datoms`, on the other hand, I see very clearly and use it very often. The norman Clojure `(reverse (datoms ...))` will return reverse iterator that would lazily walk your sequence, so I don’t think there need to be standalone `rdatoms` call.

tonsky:
(just to clarify: `datoms` is better than `seek-datoms` because it ends iteration where you expect it to end, e.g. `(datamos db :aevt :attr)` will only iterate over datoms with `:attr` attribute, wheter `seek-datoms` will finish with `:attr` and continue to different attributes)

rauhs:
Doesn't this already exist? The btset slices implement `-rseq` which gives you an efficient reverse view on the iter.

refset:
Consider the following `:eavt` index:
```
e1 :a1 v1
e2 :a1 v2
e2 :a1 v3
e2 :a2 v4
e2 :a2 v5
e2 :a2 v6
e2 :a3 v7
e3 :a4 v8
```
...my use-case is to _visually navigate_ up and down the `:eavt` index using a cursor which is able to address the individual datoms (by tracking `[:loc/e :loc/a :loc/v]`). `seek-datoms` enables me to begin iterating from the exact cursor location in the downward direction, but without `rseek-datoms` (or my equivalent workaround) I can't see a way to iterate _upwards_ that doesn't involve inefficient scanning. For instance, given my cursor is currently addressing `[e2 :a2 v5]`, I want the user to be able to "scroll-up" (visually) and see that the datom immediately above is `[e2 :a2 v4]`.

[...]

PR #254, opened 2018-02-21
Added rseek-datoms

refset:
As discussed in https://github.com/tonsky/datascript/issues/253


tonsky:
Thanks! Check out v 0.16.4

PR #255, opened 2018-03-10
fix: wrong Java equals for Entity

carocad:
Hey Nikita,
First of all thanks for this wonderful library. I started using it recently and I am loving it :)

I am using Datascript with together with some Java code (HashMaps and Heaps) for which Java uses their own `equals` method instead of Clojure's `equiv`. This of course causes several problems from my side due to two entities being equivalent but not equal (comparing memory location not values).

Let me know if I am overlooking something or if this is a quick fix as I think.

Hope it helps


tonsky:
Thanks! Should be fine, check out v 0.16.4

PR #259, opened 2018-04-19
use declare instead of defn in clojure env

shparun:
in order to avoid problems with direct linking; see issue #219


tonsky:
JFYI I’m working on a solution to this, slightly different from what you’ve proposed

tonsky:
Take a look at 0.16.5

shparun:
works now, thank you

PR #263, opened 2018-05-19
Optimize memory usage of Datom type

ronbu:
As I would like to use Datascript for large datasets, I am trying to optimise the memory usage.
This patch should reduce memory usage by at least 50%.
```clj
[tuned-datascript "0.16.5-SNAPSHOT"]
```

There are some changes to the API:
- Attributes are stored as Integers (d/datom 0 :attr :val) is not allowed anymore
- Queries return attributes as Integers, not Keywords
- You need to define the Attribute order in the schema -> {:attr {:db/order 0}}
- Transaction Id's are created from 2^24 -1 downwards
- New range for Entity ID's is from 0 to 2^24
- New range for valid Transaction ID's is from 2^24 to (2^24) - 2^20

Internally the new Datom type is implemented as a pointer to :v and a number, which stores Eid and Aid in the lower 32 bits. The remaining 21 Bits in the double number are used for the added flag and the transaction ID.

Needs some work on the JS wrapper and I could not figure out how to benchmark Datascript.

### DIFF
--- a/bench/project.clj
+++ b/bench/project.clj
@@ -1,7 +1,7 @@
 (defproject datascript-bench "0.1.0"
   :dependencies [
-    [org.clojure/clojure "1.7.0-RC1"]
-    [org.clojure/clojurescript "0.0-3308"]
+    [org.clojure/clojure "1.9.0"]
+    [org.clojure/clojurescript "1.10.217"]
     [datascript ~(or (System/getenv "BENCH_VERSION") "0.11.4")]
     [com.datomic/datomic-free ~(or (System/getenv "DATOMIC_VERSION") "0.9.5173")
       :exclusions [joda-time]]
--- a/project.clj
+++ b/project.clj
@@ -7,6 +7,8 @@
   :dependencies [
     [org.clojure/clojure "1.7.0" :scope "provided"]
     [org.clojure/clojurescript "1.7.228" :scope "provided"]
+   ;[org.clojure/clojure "1.9.0"]
+   ;[org.clojure/clojurescript "1.10.217"]
   ]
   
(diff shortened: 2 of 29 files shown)


tonsky:
Wow, lots of effort here! I don’t think I see how something like this could be merged into DataScript database. My concerns are existing clients, breaking changes, compatibility and JS portability. But I’m happy to see it exists as a separate project on its own.

Couple of questions though, purely out of curiosity:

1. Why double? Wouldn’t simply using long give you more bits to work with?

> This patch should reduce memory usage by at least 50%.

2. Have you actually measured it delivers on that goal?

>  I could not figure out how to benchmark Datascript.

Add to `profiles.clj`:

```
:bench  { :source-paths ["bench/src"] }
```

Then run (in `datascript` dir):

```
lein with-profiles +bench trampoline run -m datascript.bench/bench-all
```

ronbu:
Thanks for the feedback. I will close this PR.
If anybody is interested, my fork can be found at https://github.com/ronbu/datascript

> Why double? Wouldn’t simply using long give you more bits to work with?
It's the only number type available on JavaScript Engines.
Using two 32bit number values is another option i need to investigate. 

> Have you actually measured it delivers on that goal?
I measured memory usage in Chrome with the following code:

### Regular Datascript (167 MB)
```clj
  (def db
    (d/init-db (for [i (range 1e6)]
                (d/datom i :name i))
              {:name {:db/order 0}}))
```

### With optimized Datom type (76 MB)
```clj
(def db
    (d/init-db (for [i (range 1e6)]
                (d/datom i 0 i))
              {:name {:db/order 0}}))
```

PR #269, opened 2018-07-13
Initial set of doc strings

fversnel:
I hope the doc strings are not too long or complicated. I would really like someone to review this and suggest improvements.


tonsky:
I’m sorry, I’m in no state to do any work right now, but I’ll take a look eventually. Thanks!

PR #271, opened 2018-07-28
Add support for `:as` attr-opt; Close #270

### LINKED ISSUE #270: Support for attr-with-opts in pull syntax (:as, :default, etc.)
See https://docs.datomic.com/on-prem/pull.html#attribute-with-options
Having the ability to rename parameters and give them defaults would be very useful, and it gets us one step closer to feature parity with Datomic.

Jumblemuddle:


tonsky:
Hi, I’ll take a look eventually, maybe not soon. Sorry, and thanks!

PR #282, opened 2018-11-30
Update projects in README

CodingAgainstChaos:
We shut Bitfountain down a while ago. Just realized we're still on the README.


tonsky:
Ok, thanks for letting me know!

PR #291, opened 2019-02-06
faster (zipmap xs (range)) and has?

mpenet:
Not sure it matters that much but it's here if you need it. 
they both bench at twice the speed of the original on my machine. I only put it in v3, feel free to use it for both query namespaces (I can update the pr if needed).

---
This change is [Reviewable](https://reviewable.io/reviews/tonsky/datascript/291)

### DIFF
--- a/src/datascript/query_v3.cljc
+++ b/src/datascript/query_v3.cljc
@@ -24,6 +24,12 @@
 
 (def ^:const lru-cache-size 100)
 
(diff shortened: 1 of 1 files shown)


tonsky:
how did you measure it? jvm or js?

mpenet:
Criterium quickbench, jvm

PR #301, opened 2019-05-20
WIP: Add a test for a query that produces a `nil` :db/id

devurandom:
The attached commit adds a test that currently produces a `nil` `:db/id`, which seems wrong. Since I do not know what the correct result should be, I am treating this as work-in-progress for now.


tonsky:
The query is invalid, the problem is validation isn't picking

PR #303, opened 2019-05-27
Remove call to `persistent!` in cljs lru print function

dpsutton:
no longer pased on transients. And even if it were calling persistent
on a transient in a side-effecting operation that changes the
transient.

This error would manifest itself just by calling the lru function:

```cljs
(require '[datascript.lru :as lru])
(lru/lru 3)
**error no protocol method -persistent!

```


dpsutton:
(note this is in cljs not clj)

tonsky:
Thank you!

PR #306, opened 2019-06-03
Add `re-pattern` to builtins

dpsutton:
Clojure can resolve functions from clojure/string so that you can
check for substrings

`[(clojure.string/includes? ?term "X")]`

Clojurescript lacks this runtime resolution so there is no way to
check for string containment. There are some helpful regex functions
but this requires the regex pattern to be constructed outside of the
query. This change allows to create the pattern in the query as
demonstrated in the test:

``` clojure
(d/q '[:find  ?name
       :in    [?name ...] ?key
       :where [(re-pattern ?key) ?pattern]
              [(re-find ?pattern ?name)]]
     #{"abc" "abcX" "aXb"}
     "X")
```

### DIFF
--- a/src/datascript/query.cljc
+++ b/src/datascript/query.cljc
@@ -202,10 +202,10 @@
   'vector vector, 'list list, 'set set, 'hash-map hash-map, 'array-map array-map,
   'count count, 'range range, 'not-empty not-empty, 'empty? empty, 'contains? contains?,
   'str str, 'pr-str pr-str, 'print-str print-str, 'println-str println-str, 'prn-str prn-str, 'subs subs,
-  're-find re-find, 're-matches re-matches, 're-seq re-seq,
+  're-find re-find, 're-matches re-matches, 're-seq re-seq, 're-pattern re-pattern,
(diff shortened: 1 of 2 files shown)


tonsky:
Thanks! I also added `clojure.string/includes?` `clojure.string/blank?` `clojure.string/starts-with?` `clojure.string/ends-with?` directly to built-ins. Check out 0.18.4

dpsutton:
ah i thought about adding things like that after you accepted this PR.

I was wondering if we might modify the cljs resolve call. Currently it just returns nil but I wonder if we might add a map to expose things like clojure.string/X and clojure.set/X and any other namespaces that are always around. We can't do dynamic resolving like in clojure but there are some things that we can fake, right?

Then the clj people can just resolve as they like and the cljs people can do the same not realizing its kinda faked?

tonsky:
The main obstacle here is that with every function added to any sort of global map, it becomes inaccessible to dead code elimination. So I tend to only add what's absolute necessary. Both Clojure and ClojureScript people can always pass function reference as an additional input and call that.

dpsutton:
ah good point. I often forget about package size. thanks!

PR #318, opened 2019-09-29
extract pprinting to separate ns

zajac:
This would enable user to require namespaces behind `core` directly to eliminate startup time cost of requiring `clojure.pprint`


tonsky:
Published as 0.18.6

Issue #313, opened 2019-08-08
Implement datafy and nav for entities


thenonameguy:
This is a feature request so that entities are navigable by the new protocols introduced in Clojure 1.10. The primary use case is for REBL relation exploration.

tonsky:
PR is welcome :)

PR #325, opened 2019-10-23
Implement nav/datafy for pull

IamDrowsy:
The PR contains a version of pull/pull-many that decorates the returning maps with implementations of datafy/nav protocol.

When navigating along ref or _ref fields, it pulls the referenced entities from the db. 
Currently, when navigating, it always pulls all fields (including the reverse refs) to ensure you can navigate back and forth.

I'm not an expert in implementing datafy/nav so feedback is very welcome.

They could replace pull/pull-many of the core. As there should be no difference if you don't call datafy on the result.

For now I have not looked into ways to make the result of queries navigatable.
This touches #313 .

### DIFF
--- a/src/datascript/datafy.cljc
+++ b/src/datascript/datafy.cljc
@@ -1,0 +1,74 @@
+(ns datascript.datafy
+  (:require [clojure.core.protocols :as cp]
+            [datascript.pull-api :as dp]
+            [clojure.string :as str]))
+
+(declare datafy-entity)
+(declare datafy-entity-seq)
+
+(defn- attr<->rattr [attr]
+  (keyword (namespace attr)
+           (if (str/starts-with? (name attr) "_")
+             (subs (name attr) 1)
+             (str "_" (name attr)))))
+
+(defn- pull-pattern [ref-rattrs]
+  (into ["*"] ref-rattrs))
+
+
+(defn- navize-entity [db-val entity]
+  (let [ref-attrs (:db.type/ref (:rschema db-val))
+        ref-rattrs (set (map attr<->rattr ref-attrs))
+        many-attrs (:db.cardinality/many (:rschema db-val))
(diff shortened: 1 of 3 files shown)


tonsky:
Shouldn’t it be on entity instead? I can see how to make entity navigable, but pull results are often messy, can have multiple entites and arbitrary attrbutes mixed?

IamDrowsy:
Yeah that sounds more reasonable. I will have a look.

IamDrowsy:
I updated the PR to extend the Entity type. When datafying an entity it still uses the pull api under the hood to pull all attributes (including reverse attributes).
To use it, you currently would have to require the datascript.datafy namespace manually.

The implementation could be included into impl.entity, but this would introduce a dependency for entity to pull-api. 
So it probably the best way to make it available everywhere would be to just require datascript.datafy in datascript.core.

Btw. I also tried to have a look how datomic does implement nav/datafy (regarding what to pull etc.) on their entities, but this is not included in datomic-free.

tonsky:
Thanks! I’ll take a look later

thenonameguy:
Hey @IamDrowsy, I wanted to create a quick WIP for the same issue, your entity datafying is way better than mine, but maybe you could borrow the DB/Datom datafy code from mine:
https://github.com/tonsky/datascript/compare/master...thenonameguy:datafy_nav
Feel free to include my code.

den1k:
Merge ready?

thenonameguy:
@tonsky if you have got the time to review & merge this, it would be greatly appreciated!

tonsky:
I’ll take a look this week, sorry it was lost for so long :)

tonsky:
Sorry it took so long. The PR is fantastic, merged fully, deployed as 1.0.4. Thank you @IamDrowsy!

PR #328, opened 2019-12-09
Allow aliasing map keys in pull entry

jamesnvc:
e.g. `(d/pull ?e [{[:foo :as :bar] [:quux]}])`


tonsky:
Thanks!

tonsky:
Pushed 0.18.8

PR #332, opened 2019-12-19
`empty` preserves meta (closes #331)

### LINKED ISSUE #331: `empty` should preserve meta of db
```
(let [x (-> [1 2 3]
            (with-meta {:foo :bar})
            (empty))]
  (meta x)) ;; => {:foo :bar}        
```

https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/APersistentVector.java#L599-L601

```
(ns example
  (:require
   [datascript.core :as d]))

(let [db (-> (d/empty-db)
             (with-meta {:foo :bar})
             (empty))]
  (meta db)) ;; => nil   
```

darkleaf:


tonsky:
Thanks!

PR #337, opened 2020-02-09
list automatic ids in tempid report

### LINKED ISSUE #246: Transaction report :tempids does not always contain all tempids
If you use a map form in a transaction without including a `:db/id`, one is automatically created for the entity but it is not included in the `:tempids` key in the transaction report. Example:

```clojure
(-> (d/empty-db)
    (d/with [{:foo "a"}]))
=>
#datascript.db.TxReport{:db-before #datascript/DB{:schema nil, :datoms []},
                        :db-after #datascript/DB{:schema nil, :datoms [[1 :foo "a" 536870913]]},
                        :tx-data [#datascript/Datom[1 :foo "a" 536870913 true]],
                        :tempids #:db{:current-tx 536870913},
                        :tx-meta nil}
```

If you do the same transaction with a `:db/id` explicitly assigned, it is included:

```clojure
(-> (d/empty-db)
    (d/with [{:db/id -1
              :foo   "a"}]))
=>
#datascript.db.TxReport{:db-before #datascript/DB{:schema nil, :datoms []},
                        :db-after #datascript/DB{:schema nil, :datoms [[1 :foo "a" 536870913]]},
                        :tx-data [#datascript/Datom[1 :foo "a" 536870913 true]],
                        :tempids {-1 1, :db/current-tx 536870913},
                        :tx-meta nil}
```

If a tx-data map form does not include a `:db/id`, Datomic will automatically assign it one (like DataScript) *and* include it in the `:tempids` in the transaction report. 

Is there a reason this was implemented differently from Datomic? Would you accept a PR fixing this issue?

filipesilva:
Fix #246

### DIFF
--- a/src/datascript/db.cljc
+++ b/src/datascript/db.cljc
@@ -870,10 +870,13 @@
   [x]
   (or (and (number? x) (neg? x)) (string? x)))
 
+(defn- new-eid? [db eid]
+  (and (> eid (:max-eid db))
+       (< eid tx0))) ;; tx0 is max eid
+
 (defn- advance-max-eid [db eid]
   (cond-> db
-    (and (> eid (:max-eid db))
-         (< eid tx0)) ;; do not trigger advance if transaction id was referenced
+    (new-eid? db eid)
       (assoc :max-eid eid)))
 
 (defn- allocate-eid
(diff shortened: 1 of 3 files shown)


tonsky:
Thanks! Pushed 0.18.9

filipesilva:
Sweet, thank you very much!

PR #340, opened 2020-02-19
Issue 339: Support: db/retract without value

ericdscott:
- :db/retract with null value is treated as :db.fn/retractAttribute
- Docstring for 'transact!' is modified to reflect the change


tonsky:
Thank you! Published 0.18.10

PR #343, opened 2020-03-05
Reversed refs in query

### LINKED ISSUE #342: Reversed refs in query
I want to use reversed refs in my queries
Yes, I can replace `[?e :nested/_root ?v]` with `[?v :nested/root ?e]`, but I receive ref name as argument.

```clojure
(let [agg (-> (d/empty-db {:nested/root {:db/valueType :db.type/ref}})
              (d/db-with [{:db/id    1
                           :db/ident :root}
                          {:db/id       2
                           :nested/root :root}
                          {:db/id       3
                           :nested/root :root}]))]
  (d/q '[:find ?e ?v
         :where [?e :nested/_root ?v]]
       agg)) ;; #=> #{}, expected #{[1 2] [1 3]}
```

darkleaf:
closes #342

### DIFF
--- a/test/datascript/test/query.cljc
+++ b/test/datascript/test/query.cljc
@@ -219,5 +219,24 @@
               "X")
          #{["abcX"] ["aXb"]})))
 
+(deftest test-reverse-refs
+  (let [db  (-> (d/empty-db {:nested/root {:db/valueType :db.type/ref}})
+                (d/db-with [{:db/id    1
+                             :db/ident :root}
+                            {:db/id       2
+                             :nested/root :root}
+                            {:db/id       3
+                             :nested/root :root}
+                            {:foo :bar}]))]
+    (testing :literal
+      (is (= #{[1 2] [1 3]} (d/q '[:find ?e ?v
+                                   :where [?e :nested/_root ?v]]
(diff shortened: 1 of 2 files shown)


test/datascript/test/query.cljc
```diff
@@ -219,5 +219,24 @@
...
+                            {:db/id       2
+                             :nested/root :root}
+                            {:db/id       3
+                             :nested/root :root}
+                            {:foo :bar}]))]
+    (testing :literal
+      (is (= #{[1 2] [1 3]} (d/q '[:find ?e ?v
+                                   :where [?e :nested/_root ?v]]
+                                 db))))
+    (testing :arg
```
darkleaf: pending

darkleaf: @tonsky can you suggest solution for this case? 
I tried to add datoms with reversed ref. Also I tried to change `hash-join` fn.

darkleaf: another interesting example

```clojure
(d/q '[:find ?e ?v
       :in $ [?a ...]
       :where [?e ?a ?v]]
     db [:nested/_root :foo/bar])
```

PR #347, opened 2020-04-11
Ignore tx when diff datoms

darkleaf:


tonsky:
Thank you!

tonsky:
Pushed 0.18.11

PR #349, opened 2020-05-22
Fix `empty?` builtin

ash14:
Fixes following issue:

```clojure
(let [db (d/db-with (d/empty-db) [{:things #{"x" "y" "z"}}
                                  {:things #{}}])]
  (d/q '[:find (count ?e) .
         :where
         [?e :things ?t]
         [(empty? ?t)]]
    db))
    
 ; Expected: 1
 ; Actual: 2
```

I don't think there's any value in adding a test case just for this?

Cheers.


tonsky:
OMG I feel stupid :)) Thanks!

PR #370, opened 2020-12-08
The ability to compare values of different types (closes #369)

### LINKED ISSUE #369: Here are two cases of ClassCastException
jvm 14, clojure 1.10.1, datascipt 1.0.1

# First case

```clojure
(-> (d/empty-db {:attr {:db/cardinality :db.cardinality/many}})
    (d/db-with [[:db/add -1 :attr 1]
                [:db/add -1 :attr "str"]]))
;; BOOM
```

```clojure
(-> (d/empty-db)
    (d/db-with [[:db/add -1 :attr 1]
                [:db/add -2 :attr "str"]]))
;; it's ok
```

It fails with ClassCastException in case of  `db.cardinality/many` but works in case of `db.cardinality/one`.

# Second case

```clojure
(let [db1 (-> (d/empty-db)
              (d/db-with [[:db/add -1 :attr 1]]))
      db2 (-> (d/empty-db)
              (d/db-with [[:db/add -1 :attr "str"]]))]
  (clojure.data/diff db1 db2))

;; BOOM
```

Is PR welcome? :-)

darkleaf:


src/datascript/db.cljc
```diff
@@ -332,14 +332,20 @@
     (if (nil? y) 0
       (compare x y))))
 
+(def cljc-class #?(:cljs type :default class))
+
 (defn value-compare [x y]
```
darkleaf, line 337: @tonsky please carefully review these changes before merging

tonsky:
Comparing classes for equality won’t work. Only the class itself knows what it can compare to. E.g. PersistentArrayMap and TreeMap can be compared no problem, despite having different classes.

darkleaf:
But maps are not comparable:

```
user=> (compare {:a 1} {:a 2})
Execution error (ClassCastException) at user/eval1 (REPL:1).
class clojure.lang.PersistentArrayMap cannot be cast to class java.lang.Comparable (clojure.lang.PersistentArrayMap is in unnamed module of loader 'app'; java.lang.Comparable is in module java.base of loader 'bootstrap')
```

It seems that my patch doesn't narrow the contract, i.e. it doesn't make things worse.

tonsky:
Well, not maps, vectors, whatever. You can never know in advance. Class B can compare itself to class C, e.g. if they both are children of A. In that sense the change DOES make things worse: before, B will just compare itself to C, now it will short-circuit.

PR #372, opened 2020-12-10
Diff dbs with different types of the same attribute (closes #369)

### LINKED ISSUE #369: Here are two cases of ClassCastException

darkleaf:


tonsky:
Looks great! I think you need ClassCastException -> Exception though?

darkleaf:
No, we don't. We should notify about any exception but ClassCastException. In the JS, we can't recognize the error type so there is only `js/Error`.

tonsky:
Yeah you are right, sorry.

tonsky:
Pushed 1.0.2

Issue #378, opened 2020-12-17
Upserting with unique tuple refs fails


ash14:
Attempting an upsert to an entity using its unique tuple attribute does not work as expected.

Hope this example is clear:

```clojure
(let [conn (d/create-conn {:player {:db/unique :db.unique/identity}
                           :home {:db/valueType :db.type/ref}
                           :away {:db/valueType :db.type/ref}
                           :players {:db/unique :db.unique/identity
                                     :db/tupleAttrs [:home :away]}})]
  
  (d/transact! conn [[:db/add -1 :player "Nadal"]
                     [:db/add -2 :player "Federer"]
                     {:home -1
                      :away -2}])
  
  (d/transact! conn [{:db/id "p1"
                      :player "Nadal"}
                     {:db/id "p2"
                      :player "Federer"}
                     {:db/id "match"
                      :players ["p1" "p2"]
                      :game 3}]))

Execution error (ClassCastException) at datascript.db/value-compare (db.cljc:339).
java.lang.String cannot be cast to java.lang.Number
```

Expecting a result like `(= 3 (:game (d/entity @conn 3)))`

I don't think this is related to #364?

Thanks. I'll try to open a PR for this if it's straightforward enough.

<details>

<summary>
Edit: for what it's worth, here's the same example on Datomic 1.0.6202
</summary>

```clojure
@(d/transact conn [{:db/ident :player
                    :db/valueType :db.type/string
                    :db/unique :db.unique/identity
                    :db/cardinality :db.cardinality/one}

                   {:db/ident :home
                    :db/valueType :db.type/string
                    :db/cardinality :db.cardinality/one}

                   {:db/ident :away
                    :db/valueType :db.type/string
                    :db/cardinality :db.cardinality/one}

                   {:db/ident :players
                    :db/unique :db.unique/identity
                    :db/valueType :db.type/tuple
                    :db/tupleAttrs [:home :away]
                    :db/cardinality :db.cardinality/one}

                   {:db/ident :game
                    :db/valueType :db.type/long
                    :db/cardinality :db.cardinality/one}])

@(d/transact conn [[:db/add "p1" :player "Nadal"]
                   [:db/add "p2" :player "Federer"]
                   {:home "p1"
                    :away "p2"}])

@(d/transact conn [{:db/id "p1" :player "Nadal"}
                   {:db/id "p2" :player "Federer"}
                   {:db/id "match"
                    :players ["p1" "p2"]
                    :game 3}])

[...]

PR #379, opened 2020-12-18
Fix upserts containing tempids in tuples

ash14:
PR to address #378. Not sure how dirty the implementation is for this ... but tests pass so 🤷 

Cheers.

### DIFF
--- a/src/datascript/db.cljc
+++ b/src/datascript/db.cljc
@@ -1100,13 +1100,10 @@
                :email {\"ivan@\" 2}
                :alias {\"abc\"   3
                        \"def\"   4}}}"
-  [db entity tempids]
+  [db entity]
   (if-some [idents (not-empty (-attrs-by db :db.unique/identity))]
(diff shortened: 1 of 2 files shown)


tonsky:
Thank you for the bug report and the PR. This is an interesting problem :)

I feel that tempid resolution should not happen in `resolve-upserts`, but somewhere earlier, around https://github.com/tonsky/datascript/blob/1f060c07dfad2a700a8be702802734266dcf0222/src/datascript/db.cljc#L1431

Basically, same code should run, but for all tuple values, not just for a single v.

Also, checking for being a tuple is better not by `coll?`, but by going to schema

https://github.com/tonsky/datascript/blob/1f060c07dfad2a700a8be702802734266dcf0222/src/datascript/db.cljc#L900

ash14:
I've updated the test case - the entity to upsert now comes first which breaks the approach I was taking with this.

PR #382, opened 2021-01-10
Expose a schema as a part of the public API.

### LINKED ISSUE #381: How to get a schema correctly?
There are two ways to get a schema:

+ `(:schema db)`
+ `(datascript.db/-schema db)`

It seems these are implementation details.
Should we expose a schema via `datascript.core/schema`?
And what about `db/ref?`, `db/reverse-ref?`, `db/multival?`, etc?

darkleaf:
closes #381


tonsky:
Thank you!

Issue #373, opened 2020-12-12
Transient indices might lead to data corruption


pangloss:
Hi, I'm encountering an issue where data is disappearing or reappearing in historical versions of my db when I do subsequent transactions against the current version. It only happens with entities which have a `{:db/unique :db.unique/identity}` property.

It seems to be a data structure bug in datascript.

I can run a few transactions, query a specific instance of the db history, then do additional transactions and when I query the same instance from the history, the data will have changed.

I'm running Clojure 1.10.1 on Java 11 and have confirmed it with both v1.0.1 and v1.0.2.

I'm hoping you'll have an idea what may be causing this!

Although I can reliably reproduce it in the context of my application (it happens every time), I so far cannot figure out exactly how to reproduce it in isolation, so can't provide you a repro right now, unfortunately.

Suggestions as to how to approach reproducing the bug would be welcome. Meanwhile I think a workaround will be to isolate indexed properties into independent entities that just point to the actual data...

pangloss:
Unfortunately my hoped-for workaround didn't help, so these records are being corrupted even when they do not have unique indices attached to them, and so far I do not see anything that distinguishes them. There are only 1000-2000 entities of this type created in the transaction. Other record types are creating more in some cases without trouble. The total is about 5000 entities in the one of the larger transactions that I'm looking at.

Any change I make to the application changes the set of specific impacted entities, but if I make no changes and re-run, the results set is identical, including errors.

Here is a "small" sample that I hope shows what's happening a bit better. I'll try to explain it:

This is the code I'm using to identify bad records. In my system, every entity should have a `:type` attribute, so I can pick out records missing it.

This iterates the program one step at a time, and after each step it runs the `get-weird-ids` fn. That function goes through each historical db in the results vector and checks for entities missing the :type attribute. (An excerpt of the schema is included at the end.)

```clojure
  (defn get-weird-ids [results]
    (->> results
         (map (fn [[{:keys [db-after]}]]
                (mapv #(d/entity db-after %)
                      (distinct (map #(.e %) (d/datoms db-after :eavt))))))
         (map-indexed #(vector %1 (->> %2
                                       (filter (complement :type)) ;; every entity should have :type
                                       (map d/touch))))))

  (map second
       (reductions (fn [[results weird-ids] n]
                     ;; events/step steps the system forward and produces a single large transaction.
                     (let [results (conj results (events/step conn :tx-report))]
                       [results (get-weird-ids results)]))
[...]

PR #384, opened 2021-01-26
Filter materialized

darkleaf:
Would you like to accept this PR?
It's just a draft, if you are ok with it, I'll improve it.

Can you suggest a better implementation?

Also, I use transient, I think I should disable it due to #373.

### DIFF
--- a/src/datascript/core.cljc
+++ b/src/datascript/core.cljc
@@ -223,6 +223,9 @@
     (FilteredDB. db #(pred db %) (atom 0))))
 
 
+(def filter-materialized #'db/filter-materialized)
+
+
 ; Changing DB
 
 (defn with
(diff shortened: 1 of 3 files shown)


src/datascript/core.cljc
```diff
@@ -223,6 +223,9 @@
     (FilteredDB. db #(pred db %) (atom 0))))
 
 
+(def filter-materialized #'db/filter-materialized)
```
darkleaf, line 226: TODO: doc

test/datascript/test/filter-materialized.cljc
```diff
@@ -0,0 +1,120 @@
+(ns datascript.test.filter-materialized
```
darkleaf, line 1: TODO: remove unnecessary checks

tonsky:
Transients are fine, as long as you don’t reuse older values of them. The problem was in the logic in `transact-tx-data`, not in the transients themselves.

Is this somewhat more efficient than `(d/init-db (filter pred (d/datoms db :eavt)))`?

PR #388, opened 2021-03-05
fix: ensure transitive property for hash value comparisons

### LINKED ISSUE #386: avet index out of sync with datoms
Hi there 👋

I'm seeing some odd behaviour with Datascript where two databases with the same datoms and schema show different results for entity lookups.

For some background context, we keep a persistent log of serialized datascript transactions on a server and also save a serialized version of database every ~500 txs. Serialization used `datascript-transit`.

When the app starts, it will load the most recent serialized db and then apply all txs since then.

We noticed this problem because the we saw a serialized db render differently than the previously db plus the applied transactions.

In the code examples below, `old-db` is the db that had a series of transactions applied to it in-memory, and `new-db` is that same db after a serialization round trip.

To check these two databases were the same, I compared the datoms and schema.

```clojure
;; Datoms and schema are the same
(let [[only-in-old
        only-in-new
        in-both] (clojure.data/diff (seq old-db) (seq new-db))]
  (empty? only-in-old)       ;; true
  (empty? only-in-new)       ;; true
  (count in-both)            ;; 371900
  (= (:schema old-db) 
     (:schema new-db))       ;; true
  (identical? old-db new-db) ;; false
  )
```

I tracked it down to the entity lookup on these two databases. 

```clojure
;; old-db seems to not have this entity, but the datom exists
(let [uid "UN1JwSR7t"
      eid [:block/uid uid]
      db-id 45587
      has-uid (fn [[_ a v _]] (and (= a :block/uid) (= v uid)))
      q '[:find ?e
          :in $ ?block-uid
          :where
          [?e :block/uid ?block-uid]]]
  ;; d/entity doesn't find it on old-db
  (d/entity old-db eid) ;; nil
  (d/entity new-db eid) ;; {:db/id 45587}

  ;; datom is there
  (filter has-uid (seq old-db)) ;; (#datascript/Datom [45587 :block/uid "UN1JwSR7t" 537057490 true])
  (filter has-uid (seq new-db)) ;; (#datascript/Datom [45587 :block/uid "UN1JwSR7t" 537057490 true])

  ;; Pull can't find it either on old-db
  (d/pull old-db '[:db/id] eid) ;; throws :entity-id/missing error
  (d/pull new-db '[:db/id] eid) ;; #:db{:id 45587}

  ;; Query finds it though.
  (d/q q old-db uid)            ;; #{[45587]}
  (d/q q new-db uid)            ;; #{[45587]}
  )
```

I thought this was really weird so then I compared the indexes for a sanity check.

```clojure
;; these two indexes are the same
(let [[only-in-old
        only-in-new
        in-both] (clojure.data/diff (:aevt old-db) (:aevt new-db))]
  (count only-in-old) ;; 0
  (count only-in-new) ;; 0
  (count in-both)     ;; 371900
  )

(let [[only-in-old
        only-in-new
        in-both] (clojure.data/diff (:eavt old-db) (:eavt new-db))]
  (count only-in-old) ;; 0
  (count only-in-new) ;; 0
  (count in-both)     ;; 371900
  )

;; but this index has a bunch of entries that are different!
(let [[only-in-old
        only-in-new
        in-both] (clojure.data/diff (:avet old-db) (:avet new-db))]
  (count only-in-old) ;; 717
  (count only-in-new) ;; 177
  (count in-both)     ;; 231502 
  )
```

Worth mentioning in case it's relevant: this db includes a few `js/NaN` values that I filtered out of the results, and nested maps as values.  

I'm trying to get a standalone minimal repro up but this is not trivial since this DB contains a lot of sensitive information. I think I'm able to pare it down though.

I wanted to open the issue as soon as possible in case you had some tips on what might be happening or what I should focus on for the repro.

I couldn't find any other open issues that seemed to be the same as this one, but https://github.com/tonsky/datascript/issues/373 sounds similar.

I think this is a bug because I expect databases that have the same datoms and schema to also have the same index, and because the entity lookup should return the entity if it exists.
If there are legitimate cases that don't conform to these expectations, please let me know.

Cheers, and thanks for the awesome database!

filipesilva:
Fix #386

### DIFF
--- a/src/datascript/db.cljc
+++ b/src/datascript/db.cljc
@@ -338,8 +338,8 @@
     #?@(:clj  [(instance? Number x)       (clojure.lang.Numbers/compare x y)])
     #?@(:clj  [(instance? Comparable x)   (.compareTo ^Comparable x y)]
         :cljs [(satisfies? IComparable x) (-compare x y)])
-    #?@(:cljs [(and (or (string? x) (array? x) (true? x) (false? x))
-                 (identical? (type x) (type y))) (garray/defaultCompare x y)])
+    #?@(:cljs [(not (identical? (type x) (type y))) (garray/defaultCompare (type x) (type y))])
+    #?@(:cljs [(or (string? x) (array? x) (true? x) (false? x)) (garray/defaultCompare x y)])
     :else (- (hash x) (hash y))))
 
 (defn value-cmp [x y]
--- a/src/datascript/db.cljc
+++ b/src/datascript/db.cljc
@@ -337,7 +337,7 @@
      :cljs (identical? (type x) (type y))))
 
 (defn class-compare [x y]
-  #?(:clj  (compare (str (class x)) (str (class y)))
+  #?(:clj  (compare (.getName (class x)) (.getName (class y)))
      :cljs (garray/defaultCompare (type x) (type y))))
 
 (defn value-compare [x y]
--- a/src/datascript/db.cljc
+++ b/src/datascript/db.cljc
@@ -332,15 +332,28 @@
     (if (nil? y) 0
       (compare x y))))
 
+(defn class-identical? [x y]
+  #?(:clj  (identical? (class x) (class y))
+     :cljs (identical? (type x) (type y))))
+
+(defn class-compare [x y]
+  #?(:clj  (compare (str (class x)) (str (class y)))
+     :cljs (garray/defaultCompare (type x) (type y))))
+
 (defn value-compare [x y]
-  (cond
-    (= x y) 0
-    #?@(:clj  [(instance? Number x)       (clojure.lang.Numbers/compare x y)])
-    #?@(:clj  [(instance? Comparable x)   (.compareTo ^Comparable x y)]
-        :cljs [(satisfies? IComparable x) (-compare x y)])
-    #?@(:cljs [(not (identical? (type x) (type y))) (garray/defaultCompare (type x) (type y))])
-    #?@(:cljs [(or (string? x) (array? x) (true? x) (false? x)) (garray/defaultCompare x y)])
-    :else (- (hash x) (hash y))))
+  (try 
+    (cond
+      (= x y) 0
+      #?@(:clj  [(instance? Number x)       (clojure.lang.Numbers/compare x y)])
+      #?@(:clj  [(instance? Comparable x)   (.compareTo ^Comparable x y)]
+          :cljs [(satisfies? IComparable x) (-compare x y)])
+      (not (class-identical? x y)) (class-compare x y)
+      #?@(:cljs [(or (string? x) (array? x) (true? x) (false? x)) (garray/defaultCompare x y)])
+      :else (- (hash x) (hash y)))
+    (catch #?(:clj ClassCastException :cljs js/Error) e
+      (if (not (class-identical? x y))
+        (class-compare x y)
+        (throw e)))))
 
 (defn value-cmp [x y]
   (cond 
--- a/test/datascript/test/transact.cljc
+++ b/test/datascript/test/transact.cljc
@@ -440,3 +440,29 @@
       (is (= [(db/datom 1 :index {:map 3})]
             (vec (d/datoms db' :avet :index {:map 3} 1 )))))
 ))
+
+(deftest test-transitive-type-compare-386
+  #?(:cljs
+     (let [txs    [[{:block/uid "2LB4tlJGy"}]
+                   [{:block/uid "2ON453J0Z"}]
+                   [{:block/uid "2KqLLNbPg"}]
+                   [{:block/uid "2L0dcD7yy"}]
+                   [{:block/uid "2KqFNrhTZ"}]
+                   [{:block/uid "2KdQmItUD"}]
+                   [{:block/uid "2O8BcBfIL"}]
+                   [{:block/uid "2L4ZbI7nK"}]
+                   [{:block/uid "2KotiW36Z"}]
+                   [{:block/uid "2O4o-y5J8"}]
+                   [{:block/uid "2KimvuGko"}]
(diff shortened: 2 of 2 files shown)


tonsky:
Can you fix it for clj version too?

tonsky:
Does your test fail without the fix? I was under impression that entities were missing in broken DB, but you are checking that after deleting everything in DB it remains empty.

filipesilva:
Yes, I did verify the test fails before the fix. Verified it again just now for sanity, and that seq is `([:block/uid "wRmp6bXAx"] [:block/uid "rfL-iQOZm"])` (instead of empty) before the fix. These are the two ents that cannot be looked up due to the bug.

Regarding CLJ support, I imagine it'd look like this:
```
(defn value-compare [x y]
  (cond
    (= x y) 0
    (not (identical? (type x) (type y)))  (compare (type x) (type y))
    #?@(:clj  [(instance? Number x)       (clojure.lang.Numbers/compare x y)])
    #?@(:clj  [(instance? Comparable x)   (.compareTo ^Comparable x y)]
        :cljs [(satisfies? IComparable x) (-compare x y)])
    #?@(:cljs [(and (identical? (type x) (type y)) (or (string? x) (array? x) (true? x) (false? x))) (garray/defaultCompare x y)])
    :else (- (hash x) (hash y))))
```

WDYT?

tonsky:
1. I think we should replace type with class (`type` returns something from metadata or class, `class` is always a class).
2. Figure out how to compare classes (they are not comparable).
3. Type comparison should go after `Comparable` check, otherwise we’ll fail to compare Int and Long, Vector and Seq etc

```
user=> (compare (type 1) (type "a"))
Execution error (ClassCastException) at user/eval15 (REPL:1).
class java.lang.Class cannot be cast to class java.lang.Comparable (java.lang.Class and java.lang.Comparable are in module java.base of loader 'bootstrap')
```

filipesilva:
1. Will do class for CLJ, but CLJS doesn't have that fn.
2. WDYT of `(compare (str (class x)) (str  (class y)))`?
3. There might be some extra subtlety there, because string is comparable but will throw when compared to number
```
(instance? Comparable "abc")
true
(.compareTo "abc" 3)
Execution error (ClassCastException) at user/eval17 (REPL:1).
class java.lang.Long cannot be cast to class java.lang.String (java.lang.Long and java.lang.String are in module java.base of loader 'bootstrap')
```

filipesilva:
Another problematic case, this one for CLJS
```
(satisfies? IComparable [1 2 3]) ;; => true
(-compare [1 2 3] 2) ;; => error, cannot compare
```

filipesilva:
Push a change that checks for different classes after the comparable check, but also catches errors and checks for different classes there too. LMK what you think.

tonsky:
Judging from http://www.docjar.com/html/api/java/lang/Class.java.html, Class toString allocates new string, which is slow. Better use Class.getName, which doesn’t allocate.

As for CLJS, I wonder what type result of `type` is? But there’s `type->str`, so maybe we can just use that? https://github.com/clojure/clojurescript/blob/69b66374d857a176c5a9cd8cc0cbb7506876640d/src/main/cljs/cljs/core.cljs#L328-L331

Otherwise, looks good. Nice touch on try/catch!

filipesilva:
On CLJS `type` will return the constructor itself, so I'm kinda wary of using `type->str` because I (maybe?) makes a new string, but at any rate two fns with the same name and body will be the same string, which sounds wrong. `garray.defaultCompare` will just really call `>` and that seems to compare the fns in a consistent way anyway, so I used it. If you'd still like me to use `type->str` though, happy to oblige.

For CLJ it does sound better to use get name, will change that.

tonsky:
Looks good. I’m curious about function comparsion, asked here https://twitter.com/nikitonsky/status/1367984972519313410?s=20. Probably by tomorrow we’ll have an answer, I’ll merge it then.

Thanks for the great work and contribution! I’m glad we got to the bottom of this.

tonsky:
Ok, seems like comparing functions is [going through `toString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/toString)

One problem here is that `==` doesn’t go through toString, but `<` `>` `<=` `>=` do: 

```
> let a = () => 1;
undefined

> let b = () => 1;
undefined

> a < b
false

> b < a
false

> a <= b
true

> b <= a
true

> a == b
false

> a === b
false
```

Another is that in some cases functions return `[native code]` as their body, makign toString comparison return false positive:

```
> (function f(){ return 1; }.bind(0)).toString()
"function() {
    [native code]
}"

> (function f(){ return 2; }.bind(0)).toString()
"function() {
    [native code]
}"

> (function f(){ return 1; }.bind(0)) < (function f(){ return 2; }.bind(0))
false

> (function f(){ return 1; }.bind(0)) > (function f(){ return 2; }.bind(0))
false

> (function f(){ return 1; }.bind(0)) == (function f(){ return 2; }.bind(0))
false
```

I think we can go through `(garray/defaultCompare (type->str (type a)) (type->str (type b)))`. It will make functions returning `[native code]` being equal. That’s unfortunate, but I don’t see how we can’t work around it.

I’ll add the change and merge the PR now.

tonsky:
Merged and released 1.0.6. Thanks again!

tonsky:
Oops, I mean 1.0.7 :)

filipesilva:
I **think** the end result of this JS behaviour is not a problem, because we only compare types when they are not identical, and identical doesn't seem to suffer from the problem of `==` as you described. 

Which is to say, when we compare we already know they are not identical, and JS may give false positives ("the comparator ret for these two is 0") on the >/< check, but it doesn't give false negatives ("this one is bigger than that one"). Which, as I'm writing this I realise, means it could lead to things not being inserted in the index at all.

Not sure what the best thing to do in JS would be here really. But I think these changes are a step forward regardless.

Issue #394, opened 2021-05-19
Integrated Datalog Admin Console?


groundedSAGE:
There is now a Chrome Extension and plans for an Electron application for the [Datalog Console](https://github.com/homebaseio/datalog-console).

Our thoughts were to have this integrated with Datascript. @tonsky what would this integration look like for you?

PR #395, opened 2021-05-19
feat(datalog-console): respond to datalog-console messages in browser

groundedSAGE:
Basic implementation for Datalog Console #394

### DIFF
--- a/src/datascript/core.cljc
+++ b/src/datascript/core.cljc
@@ -2,6 +2,7 @@
   (:refer-clojure :exclude [filter])
   (:require
     [#?(:cljs cljs.reader :clj clojure.edn) :as edn]
+    [datascript.integrations.datalog-console :as datalog-console]
     [datascript.db :as db #?@(:cljs [:refer [FilteredDB]])]
     #?(:clj [datascript.pprint])
     [datascript.pull-api :as dp]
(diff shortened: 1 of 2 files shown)


tonsky:
Not saying your approach is wrong, just need more context and make sure we looked at all options. This is what I am curious about:

1. Is there a way for initialization code to not be in the core?
2. You are hooking up conn creation. What do you think should happen if user has multiple conns?
3. Same question, but if user just uses db and db-with?

groundedSAGE:
Glad you like the Datalog Console. 

We are aware that there is quite a few things we need to consider and this is just to get the ball rolling. These are of course very good questions. 

I think we could move the code outside of core and have some utility for registering the database. Not sure exactly what that would look like yet. We are open to suggestions on what could be done here if you have any.

tonsky:
What if you just insert something like

```
datalog_console.show_db(db);
```

? What would be limitations of that? Usability-wise, convenience-wise?

groundedSAGE:
We are currently doing a lot of improvements to the Datalog Console and it will now be available as a separate library. I am closing this pull request and will open another one for a README update when v1 is released. So developers can find dev tooling that works with Datascript.

PR #397, opened 2021-07-06
fix #315: lookup ref should not throw in pull

### LINKED ISSUE #315: Exception thrown when lookup ref used in pull
When I `d/pull` an entity using a lookup ref, an exception is thrown if the entity does not exist. This behavior does not match the behavior of `d/pull` when passed a `:db/id`. Example cases:

`pull`'ing an entity that does not exist using a db/id.
```clojure
(d/pull (d/empty-db {:id {:db/unique :db.unique/value}})
        [:foo]
        2)
=> nil
```

`pull`'ing an entity that does not exist using a lookup ref: 
```clojure
(d/pull (d/empty-db {:id {:db/unique :db.unique/value}})
        [:foo]
        [:id "a"])
Error: Nothing found for entity id [:id "a"]
```

Is there a reason the behavior is different? FWIW, Datomic does not throw an exception when `pull`'ing an entity that does not exist via lookup ref.

timothypratley:
See https://github.com/tonsky/datascript/issues/315

A pull ref lookup failure should result in nil rather than an error.


tonsky:
Thanks for getting to the bottom of this! Would you be so kind adding a simple test for this, too?

timothypratley:
@tonsky good thinking! Indeed there was a problem (returning too many results instead of on nil - fixed), and also another bug: If there were no results for a pull-many, it was being omitted from the result vector instead of being nil (see the test cases if this isn't clear) - also fixed.
It has also revealed other inconsistencies... so this requires some more work still; I'll let you know when those are investigated.

timothypratley:
@tonsky O.K. this is ready for your review now.

tonsky:
Wow thanks for setting up the infrastructure! I’ll take a look at it next week

timothypratley:
You're welcome. Have a great weekend. 😄

tonsky:
Published as 1.2.3

PR #408, opened 2021-10-31
Allow :db/valueType :db.type/tuple

### LINKED ISSUE #407: :db/valueType :db.type/tuple support in schema?
I'm working on some code that uses composite tuples. Looking at the datomic documentation, I was trying to define such a tuple as

```
{:db/valueType :db.type/tuple
  :db/tupleAttrs [:my/att1 :my/att2]
  :db/unique :db.unique/identity}
```
But that fails with

```
Bad attribute specification for ..., expected one of #{:db.type/ref}
```
Reading the datascript docs more closely it seems that I *must not* use db/valueType here; composite attributes are defined by having a  :db/tupleAttrs attribute in their schema.

Is this difference in behavior intentional? I wouldn't mind providing a patch to allow a db.type/tuple in this specific case if that's a welcome addition.

joodie:
Allow specifying composite tuple with :db/valueType :db.type/tuple

This schema attribute is optional, but when specified will check that
:db/tupleAttrs is also provided.

This fixes #407

### DIFF
--- a/src/datascript/db.cljc
+++ b/src/datascript/db.cljc
@@ -728,8 +728,16 @@
            :key       :db/isComponent})))
 
     (validate-schema-key a :db/unique (:db/unique kv) #{:db.unique/value :db.unique/identity})
-    (validate-schema-key a :db/valueType (:db/valueType kv) #{:db.type/ref})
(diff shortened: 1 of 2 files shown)


tonsky:
Looks great, thank you! Published as 1.2.9

PR #417, opened 2021-12-27
Allow symbols as constants

### LINKED ISSUE #416: Symbols are not allowed as constants
With Datomic this query parses and returns `#{[1]}`:
```
(q
    '[:where [?e :s plain]
      :find ?e]
    '[[1 :s plain]])
```
treating `plain` as a constant. This fits with Datomics documentation that defines a constant as "any non-variable data literal". Datascript fails to parse the query as it doesn't recognise the data pattern.

Note that source variables(those that start with $) not in first position are also treated as constants so the following also parses and runs in Datomic
```
(q
    '[:where [?e :s $plain]
      :find ?e]
    '[[1 :s $plain]])
```

I expected Datascript to return the same result as Datomic in this case.

I have a minor update to the parser that fixes this and some tests I can contribute.

zoren:
Closes #416. 

This PR allows using symbols(not starting with `?`) as constants in data patterns. This is consistent with Datomic's behaviour. 

Notes:

- I changed the order of parsers in `parse-fn-arg` otherwise `$x` would parse as a constant, and it shouldn't for function arguments.

- I believe the unit test saying `(friends something)` should throw is wrong. I've changed it to a positive test where `something` parses as a constant argument instead.

### DIFF
--- a/src/datascript/parser.cljc
+++ b/src/datascript/parser.cljc
@@ -116,8 +116,9 @@
     (RulesVar.)))
 
 (defn parse-constant [form]
-  (when (not (symbol? form))
-    (Constant. form)))
+  (when-not (and (symbol? form)
+                 (= (first (name form)) \?))
+   (Constant. form)))
 
 (defn parse-plain-symbol [form]
   (when (and (symbol? form)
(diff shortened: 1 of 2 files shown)


tonsky:
Thank you! Released 1.3.3

zoren:
> Thank you! Released 1.3.3

Thank you for making datascript!

PR #418, opened 2021-12-29
Render all variables in rule arity error message

zoren:
Currently `flatten-rule-vars` will only return the free variable symbols if there are required variables. This is due to a misplaced `)`. This is not a big issue as `flatten-rule-vars` is only called in to generate an error message in rule arity validation.


tonsky:
Thank you!

PR #421, opened 2022-01-17
Comparison of non numbers

### LINKED ISSUE #420: Comparison predicates do not work for strings
Datomic supports comparisons of strings in predicates like so:
```
[:find ?name
 :where [_ :artist/name ?name]
        [(<= "Q" ?name)]
        [(< ?name "R")]]
```
Should we also support this in datascript?

zoren:
Closes #420.

Uses `datascript.db/value-compare` in query comparison predicates to allow comparison of non-numbers.

Note: This changes comparison predicates to only support arity two as in Datomic.


tonsky:
Thanks! I added multi-arities just for fun, and released 1.3.6

zoren:
> Thanks! I added multi-arities just for fun, and released 1.3.6

Great! Good Clojure-spirit to have it be a non-breaking update!

Issue #360, opened 2020-07-12
feature-request: support for arbitrary query functions (besides built-ins) through an `:allow` list


peey:
Currently, I believe only functions in https://github.com/tonsky/datascript/blob/master/src/datascript/query.cljc#L194 are allowed to appear in queries.

Would it be possible to have a datomic-like feature for an `:allow` list for whitelisting other custom functions https://docs.datomic.com/cloud/ions/ions-reference.html#ion-config ?

peey:
My bad, it seems like passing any FQNs to datascript in place of a built-in already works!

tonsky:
Yes. And you can also pass any functions as input and then call them:

```
(d/q '[:find ?x
       :in $ ?inc
       :where [?e :attr true]
              [(?inc ?e) ?x]]
  db #(+ % 1))
```

PR #423, opened 2022-01-31
Add get to built-ins for binding on map values

logseq-cldwalker:
Hi again! Would you be open to adding `get` to the builtins? It's the only thing keeping us @logseq from getting off of our datascript fork. I'm aware we could pass `get` as [an input](https://github.com/tonsky/datascript/issues/360) but are needs are a bit complex as our users already rely on this builtin to writing datalog queries in our editor. I'd also love for logseq to be using standard datascript so we can start to explore using datascript via bb. Thanks for this library as it has done wonders for logseq! Cheers

### DIFF
--- a/src/datascript/built_ins.cljc
+++ b/src/datascript/built_ins.cljc
@@ -78,9 +78,9 @@
      false)))
 
 (def query-fns {
-  '= =, '== ==, 'not= not=, '!= not=, 
-  '< less, '> greater, '<= less-equal, '>= greater-equal, 
-  '+ +, '- -, '* *, '/ /, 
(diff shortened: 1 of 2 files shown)


logseq-cldwalker:
While writing the test I noticed the test instructions didn't work. Locally I added the necessary kaocha deps and tests.edn to test this PR. I could add these changes to this PR or a separate one if that's helpful

tonsky:
> While writing the test I noticed the test instructions didn't work

That’s ok, I’ll clean it up eventually

logseq-cldwalker:
Thanks!

tonsky:
Pushed 1.3.8

PR #426, opened 2022-02-16
Allow symbols as data in where patterns

### LINKED ISSUE #425: Symbols not compared as expected in queries
The query:
```
(d/q
   '[:find [?e ...]
     :where [?e :s b]]
   '[[1 :s a]
     [2 :s b]])
```
returns `[2 1]` in datascript but I would expect `[2]` which is also what I get with Datomic.

zoren:
Closes #425

When a symbol in a data pattern is not `_` or starts with `?` we should consider it data.


tonsky:
Thank you!

tonsky:
Published 1.3.10

PR #446, opened 2023-02-24
Allow the use of lookup refs in get-else and get-some

### LINKED ISSUE #445: `get-else` fails when passed a lookup.
The datomic docs suggest [any entity identifier is allowed](https://docs.datomic.com/cloud/query/query-data-reference.html#get-else) for the entity argument passed to `get-else`. However in datascript an exception is thrown when a lookup vector is used.
```clojure
(d/q
    '[:find ?maybe
      :in $ ?lookup
      :where [(get-else $ ?lookup ::number 0) ?maybe]]
    (d/db-with (d/empty-db {::id {:db/unique :db.unique/identity}})
      [{::id 123}])
    [::id 123])
```
Results in:
```
 ClassCastException class clojure.lang.PersistentVector cannot be cast to class java.lang.Number
```

jjttjj:
Fixes #445

### DIFF
--- a/src/datascript/built_ins.cljc
+++ b/src/datascript/built_ins.cljc
@@ -12,7 +12,7 @@
   [db e a else-val]
   (when (nil? else-val)
     (raise "get-else: nil default value is not supported" {:error :query/where}))
-  (if-some [datom (first (db/-search db [e a]))]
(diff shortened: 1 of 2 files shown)


tonsky:
Thanks!

tonsky:
Pushed 1.4.1

Issue #447, opened 2023-03-20
`entity` returns entities for ids that don't exist


lynaghk:
The docs suggest that `entity` should return `nil` when the provided id doesn't actually correspond to an entity in the database:

https://github.com/tonsky/datascript/blob/1a405721cb2226a28678066b176a5326c264153e/src/datascript/core.cljc#L31-L33

but in fact it always returns an entity. E.g., with datascript 1.4.1 on JVM Clojure:

```clojure
(d/entity (d/empty-db) 100500) ;; => #:db{:id 100500}
```

The only check seems to be that the number isn't above `emax`

https://github.com/tonsky/datascript/blob/1a405721cb2226a28678066b176a5326c264153e/src/datascript/db.cljc#L1042-L1044

I would prefer the behavior described in the docs, which is useful when "rebasing" entities by their id across different databases (i.e., one can use `keep` or other Clojure nil punning fns to discard any entities that may have been removed in a derived database).

Happy to submit a patch if maintainers agree that the documented behavior is, in fact, the desired one.

refset:
Arguably related: https://github.com/tonsky/datascript/issues/187

logseq-cldwalker:
Looks like this was fixed with [1.4.2](https://github.com/tonsky/datascript/blob/master/CHANGELOG.md#142). Should this be closed?

PR #448, opened 2023-03-20
Entity should return nil for numeric eids that don't correspond to entities

lynaghk:
Ref: https://github.com/tonsky/datascript/issues/447

I'm not sure about the fastest way to check that an entity exists in the database.
I used `(empty? (db/-search db [e]))` here, but presumably there's some allocation overhead from creating an iterator.

Perhaps there's a cheaper way to do it (maybe seeking on the index directly and checking for an eid match?)
I'm open to suggestions and am happy to take a second look later with profiling if desired.

### DIFF
--- a/src/datascript/core.cljc
+++ b/src/datascript/core.cljc
@@ -75,9 +75,7 @@
 (def ^{:arglists '([db eid])
        :doc "Given lookup ref `[unique-attr value]`, returns numberic entity id.
 
-             If entity does not exist, returns `nil`.
-
-             For numeric `eid` returns `eid` itself (does not check for entity existence in that case)."}
+             If entity does not exist, returns `nil`."}
   entid db/entid)
 
 
--- a/src/datascript/db.cljc
+++ b/src/datascript/db.cljc
@@ -1066,6 +1066,9 @@
     (raise "Expected number or lookup ref for entity id, got " eid
       {:error :entity-id/syntax, :entity-id eid})))
 
+(defn numeric-eid-exists? [db eid]
+  (= eid (-> (-seek-datoms db :eavt [eid]) first :e)))
+
 (defn entid-strict [db eid]
   (or (entid db eid)
       (raise "Nothing found for entity id " eid
--- a/src/datascript/impl/entity.cljc
+++ b/src/datascript/impl/entity.cljc
@@ -14,7 +14,8 @@
 (defn entity [db eid]
   {:pre [(db/db? db)]}
   (when-let [e (entid db eid)]
-    (->Entity db e (volatile! false) (volatile! {}))))
+    (when (db/numeric-eid-exists? db e)
+      (->Entity db e (volatile! false) (volatile! {})))))
 
 (defn- entity-attr [db a datoms]
   (if (db/multival? db a)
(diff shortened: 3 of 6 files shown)


lynaghk:
Seems like there's a bit more work to be done based on the failing JavaScript test here: https://github.com/tonsky/datascript/blob/cb06ac762e2f99d30bdc8fb09b49709eac5e5999/test/js/tests.js#L289

caused by the 101 not resolving to anything.
I.e., there is no `{:db/id 101}` entity created when the numeric 101 is added as a child reference here: https://github.com/tonsky/datascript/blob/cb06ac762e2f99d30bdc8fb09b49709eac5e5999/test/js/tests.js#L283

There's a similar CLJ test, but it passes because all entities are defined at the toplevel https://github.com/tonsky/datascript/blob/cb06ac762e2f99d30bdc8fb09b49709eac5e5999/test/datascript/test/entity.cljc#L49-L52

Note there's one failing JavaScript test here: https://github.com/tonsky/datascript/blob/cb06ac762e2f99d30bdc8fb09b49709eac5e5999/test/js/tests.js#L289

caused by the 101 not resolving to anything.
I.e., there is no `{:db/id 101}` entity created when the numeric 101 is added as a child reference here: https://github.com/tonsky/datascript/blob/cb06ac762e2f99d30bdc8fb09b49709eac5e5999/test/js/tests.js#L283

There's a similar CLJ test, but it passes because all entities are defined at the toplevel https://github.com/tonsky/datascript/blob/cb06ac762e2f99d30bdc8fb09b49709eac5e5999/test/datascript/test/entity.cljc#L49-L52

What's the desired behavior for transactions with references to positive numeric IDs that don't correspond to entities? For maps it's pretty clear the entities should be created. But for numeric IDs, I'm not sure. (Especially not sure for positive numbers, since `tempid` always returns negative ones.)

tonsky:
> I'm not sure about the fastest way to check that an entity exists in the database.

Interesting question! I’d go with something like

```
(= eid (-> (-seek-datoms db :eavt eid) first :e))
```

Might be a bit faster because it only does one binary search for left boundary. For some reason I thought there was some obvious easy way to check for entity, but I guess there isn’t. Maybe that’s why Datomic does not check eid, too.

> What's the desired behavior for transactions with references to positive numeric IDs that don't correspond to entities? For maps it's pretty clear the entities should be created. But for numeric IDs, I'm not sure. (Especially not sure for positive numbers, since tempid always returns negative ones.)

Interesting question! I guess we want to throw, but then it’ll be a breaking change... Also should make sure that forward-referencing something still works with tempids (clauses are resolved one-by-one, and tempids replaced with allocated new eids, hopefully won’t be a problem) like this:

```
[[:db/add -1 :friend -2]
 [:db/add -2 :friend -1]]
```

Hopefully it’ll be ok to throw in this case though (both 1 and 2 do not exist yet). If you want this, use tempids:

```
[[:db/add 1 :friend 2]
 [:db/add 2 :friend 1]]
```

lynaghk:
I had a go at throwing when positive IDs which don't correspond to entities are used in transactions, but this turned out to be pretty tricky --- it seems like a fair bit of the code right now relies on the behavior that positive IDs will pass through `entid` and `entid-strict` even if they don't correspond to entities that exist.

To keep this PR small, I just fixed the public API `entity` behavior and the (arguably broken) JS tests which relied on the above behavior.

Let's revisit the question of positive IDs in a separate issue / PR, since I suspect it'll require reworking a lot of internals.

tonsky:
Thanks! Merged and published as 1.4.2. Let’s hope world will survive :)

PR #453, opened 2023-08-04
Make issue refs clickable in README

igoradamenko:
Hey @tonsky.

I was just passing by and noticed that issue refs are not clickable. Fixed that.


tonsky:
Thanks!

Issue #456, opened 2023-11-15
Query performance with rules is much worse than "equivalent" inline clause


RutledgePaulV:
I was attempting to clean up a few queries by factoring out some simple rules from my `:where` clauses. However, I immediately began noticing a big decrease in query performance and increased memory usage. I was able to simplify it down to a pretty simple example.

I reproduced the issue on both 1.4.2 and 1.5.4 so I don't think it's a recent regression.

```clojure
(require '[datascript.core :as d])

(def db
    (-> (d/empty-db)
        (d/db-with (for [x    (range 50000)
                         :let [temp (str (random-uuid))]
                         fact [[:db/add temp :item/id x]
                               [:db/add temp :item/status (rand-nth ["started" "pending" "stopped"])]]]
                     fact))))

; inline clauses, this is fast / performs as expected
(time
    (def results
      (d/q '[:find ?e
             :where
             [?e :item/status ?status]
             [(ground "pending") ?status]]
           db)))

; "Elapsed time: 13.563333 msecs"


; logically equivalent rule, but much slower and sometimes exceeds my default max heap!
(time
    (def results
      (d/q '[:find ?e
             :in $ %
             :where
             [?e :item/status ?status]
             (pending? ?status)]
           db
           '[[(pending? ?status)
              [(ground "pending") ?status]]])))

; "Elapsed time: 53745.881542 msecs"
```

tonsky:
Closed in #457

PR #457, opened 2023-11-15
prevent explosion of redundant tuples during rule solving

RutledgePaulV:
Addresses performance problem reported in https://github.com/tonsky/datascript/issues/456

Prior to this change `tuples` contained a bunch of duplicates due to calling `-collect` instead of `collect` which does deduplication. Those duplicates then propagated across the relations causing an unnecessary amount of work.

### DIFF
--- a/src/datascript/query.cljc
+++ b/src/datascript/query.cljc
@@ -609,7 +609,8 @@
 
             ;; no rules -> expand, collect, sum
             (let [context (solve (:prefix-context frame) clauses)
-                  tuples  (collect context final-attrs)
+                  tuples  #?(:clj (collect context final-attrs)
+                             :cljs (-collect context final-attrs))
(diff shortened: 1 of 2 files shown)


RutledgePaulV:
In the latest commit I descoped the fix for ClojureScript and because it's more complex. All tests now pass on both platforms.

I'll propose we apply the issue on the JVM first since it's pretty straightforward and perhaps @tonsky can help with the ClojureScript side since it's non-obvious to me why the same fix doesn't work for both. I could see an argument for "fix the compatibility issues" so that the same performance fix works for both, but there may be good reasons for why they're divergent and need separate solutions.

tonsky:
I think cljs was not expected all tuples (arrays) to be converted to vectors just yet. Thanks for the contribution! Pushed 1.6.0

tonsky:
1.6.1 :)

PR #459, opened 2024-02-13
Query: shortcircuit query when result is guaranteed to be empty

galdre:
The `-collect` fn, run at the end of a query, correctly shortcircuits with the following comment: `"one empty rel means final set has to be empty"`

However, while processing the query clauses, there was no shortcircuit. Because the implementation of `lookup-pattern` is agnostic to existing rels, this means that the query engine will continue slicing the indexes for every tuple that could possibly match every clause until all clauses are exhausted.

This is a one-line change to add a shortcircuit in `resolve-clause` if any existing relation is empty. The check is cheap, and it should provide a substantial performance boost in many common cases.

Motivation: substantial performance boost for a real-world app

### DIFF
--- a/bench/datascript/bench/datascript.cljc
+++ b/bench/datascript/bench/datascript.cljc
@@ -110,6 +110,19 @@
                   [?e :sex :male]]
       @*db100k)))
 
+(defn bench-q5-shortcircuit []
+  (bench/bench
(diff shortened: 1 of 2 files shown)


tonsky:
Awesome! Thank you!

tonsky:
Published 1.6.2

PR #460, opened 2024-02-14
Bugfixes for sum-rel and shortcircuited or/or-join branches

galdre:
[My previous PR](https://github.com/tonsky/datascript/pull/459) inadvertently broke or/or-join when a branch shortcircuits due to being empty. (Sorry!)

If one branch of an `or` or `or-join` short-circuited, the resulting Relation would not have the expected attrs, resulting in `Can’t sum relations with different attrs` being raised.

This fixes the bug and adds regression tests. Fixing that also exposed an extant bug in `same-keys?` which is now fixed.

### DIFF
--- a/src/datascript/query.cljc
+++ b/src/datascript/query.cljc
@@ -133,7 +133,9 @@
       (= attrs-a attrs-b)
       (Relation. attrs-a (into (vec tuples-a) tuples-b))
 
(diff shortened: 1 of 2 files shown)


tonsky:
I though of that but hoped we already have tests. Good catch!

PR #465, opened 2024-04-18
[clj] serialize bigdecs as other type, for cross-platform compatibility.

### LINKED ISSUE #466: Handling of BigDecimal
I wanted to add further clarification to the importance of #465, because I realized there was not already an open issue about it.

We have an application where the server (clj) wants to send a serialized datascript database to the client (cljs).  The database contains values that are BigDecimals.

Currently, datascript treats `number?` [the same way in clj and cljs](https://github.com/tonsky/datascript/blob/826781bc0348590f3fc5c9013321d8af25a4a989/src/datascript/serialize.cljc#L142), but clojurescript has no concept of a BigDecimal.

#465 allows the `freeze-fn` and `thaw-fn` to handle BigDecimals.  In our case, we have a cljs thaw-fn that decodes BigDecimal into [decimal.js](https://github.com/MikeMcl/decimal.js) types, and this is working for us in our app.

aiba:
This is required if you have a server (clj) serialized datascript DB that is deserialized by a client (cljs).  Since cljs does not have a notion of bigdec, it has to be handled specially be the `thaw-fn`.  Fixes #466

### DIFF
--- a/src/datascript/serialize.cljc
+++ b/src/datascript/serialize.cljc
@@ -138,6 +138,7 @@
                       (cond
                         (string? v)  v
                         #?@(:clj [(ratio? v) (write-other v)])
+                        #?@(:clj [(decimal? v) (write-other v)])
                         
                         (number? v)  
                         (cond


tonsky:
Thanks! Done

Issue #462, opened 2024-04-16
Index usage doesn't match datomic behavior and appears to be much less efficient


latacora-paul:
Hi! 

I've started measuring how many datoms are "scanned" by a query using db filters and was surprised to see way more datoms being considered than I expected to answer pretty simple queries. I then checked this behavior against datomic and saw that the number of datoms being considered there exactly matched my mental model for how queries translate into index traversals. 

This suggests to me that there are some sizable performance gains for certain queries if we can figure it out. This is especially relevant to me because I'm trying to use the new storage protocols against a remote store (dynamodb) and need to minimize the amount of load calls.

Below you'll find two tests that demonstrate the behavior on each platform - in the datascript case it's considering a total of 50 datoms to answer the query (even though there are only 25 in the db!) whereas in datomic it's only looking at 2 (as expected).

This was tested on version 1.6.3 of datascript and version 1.0.6733 of datomic peer.

Datascript:

```clojure
(deftest datascript-example
  (let [count-datoms-scanned
        (fn [db query & args]
          (let [inspected-datoms (atom [])
                filtered-db      (datascript.core/filter db (fn [db datom] (swap! inspected-datoms conj datom)))
                answer           (apply datascript.core/q query filtered-db args)]
            [(count (deref inspected-datoms)) answer]))
        letters
        (map (comp keyword str char) (range 97 123))
        schema
        (reduce (fn [schema letter] (assoc schema letter {:db/index true})) {} letters)
        tx-data
        (for [[pre post] (partition 2 1 letters)]
          [:db/add (str (random-uuid)) pre post])
        db
        (-> schema
            (datascript.core/empty-db)
            (datascript.core/db-with tx-data))
        magic-entity-id 8
        magic-attribute-id :h]
    ; THIS FAILS - DATASCRIPT IS ACTUALLY LOOKING AT 50 DATOMS INSTEAD OF EXPECTED 2
    (is (= [2 #{[magic-entity-id magic-attribute-id :i]}]
           (count-datoms-scanned
             db
             '[:find ?e ?a ?v
               :in $ ?entity
               :where
               [?entity ?a ?v]
               [?e ?a ?v]]
             magic-entity-id))
        "I can jump straight to the datoms that matter because of indices")))
```

Datomic:
```clojure
; PS the "magic-entity-id" was constant in my trials but I'm not sure how well it translates to other machines/versions

(deftest datomic-example
  (let [count-datoms-scanned
                           (fn [db query & args]
[...]

PR #467, opened 2024-04-24
treat a query as a progressive funnel instead of independent clauses to search and join

latacora-paul:
Hey!

Let's consider this a work in progress since I broke something around lookup refs (will look into it) and I know you might want to make some different implementation decisions. However, the query tests are passing and this greatly reduces the number of datoms being fetched from storage in many cases.

Your fix for https://github.com/tonsky/datascript/issues/462 addressed the "constant / one possibility" case but it didn't do anything to constrain the search space in most other cases (like when there are only a few possible values for a binding as determined by an earlier where clause). 

Here I'm constraining the search space of where clauses using candidate binding values (which considerably decreases the number of datoms that need to be scanned for clauses appearing later in a query). In other words, this handles both the "constant" case and allows relations from earlier where clauses to turn later searches into "a collection of constant cases".

### DIFF
--- a/src/datascript/query.cljc
+++ b/src/datascript/query.cljc
@@ -831,7 +831,6 @@
                  (sum-rel relation (Relation. bindings (:tuples rel')))))
              (Relation. bindings [])
              bound-patterns)]
-       (println "relation" relation)
        (binding [*lookup-attrs*
                  *lookup-attrs*]
          (update context :rels collapse-rels relation))))))


tonsky:
If I understand your approach right:

- First clause does some basic filtering, e.g. `[10 ?a ?v]` produces some relation with N amount of `[?a ?v]` pairs. E.g. `[:name 1] [:name 2] [:age 3] [:age 4]`
- If second clause has either `?a` or `?v`, you will run N small queries against database. E.g. if second clause is `[?e :attr ?v]`, you will run separate `lookup-pattern` for `?e :attr 1`, `?e :attr 2`, `?e :attr 3` and `?e :attr 4`

I agree that this is more precise and will fetch less datoms. I am not sure about the performance of it though. What if first clause returned 100 possible values for `?v`? It might be that just fetching all `[?e :attr ?v]` and then doing a hash-join might be faster? Maybe we need a benchmark to prove it works?

I am also not sure where to draw the line. Sure, for one value of `?v`, adding it to the pattern is faster. For maybe 2-3, too. For 100, not so sure. When to stop?

Do you have any insight into how Datomic does it? Or XTDB maybe?

galdre:
I have also experimented with a similar patch to this one, and found that it was sometimes much faster and sometimes far slower for my app; I also experimented with inserting an arbitrary threshold, but overall I strongly suspect (partly due to experience with [eva](https://github.com/Workiva/eva)):

1. It will greatly depend on the shape and scale of your data.
2. Performance characteristics will dramatically differ based on whether or not you are using persistence (not to mention the latency of the persistence).

One low-hanging approach that might be beneficial to use cases like those of @latacora-paul could be to switch strategies not based on a threshold, but simply based on whether or not the data is persistently stored.

Another approach I've been mulling is to use Clojure/script's metadata to be able to hint to the query engine specific strategies to override defaults.

Issue #474, opened 2024-06-20
:db.fn/call returning entity without :db/id fails (regression in 1.7.0)


DerGuteMoritz:
Returning an entity from a `:db.fn/call` without an explicit `:db/id` doesn't work anymore since 1.7.0 but did still work in 1.6.5.

That is, the following test:

```clojure
(deftest test-entity-returned-from-db-fn-without-id
  (let [conn   (d/create-conn {})
        _      (d/transact! conn [[:db.fn/call (fn [db]
                                                 [{:foo "bar"}])]])
        db     @conn]
    (is (= #{[1 :foo "bar"]} (tdc/all-datoms db)))))
```

Passes on 1.6.5 but fails on 1.7.0 with the following error:

```
ERROR in (test-entity-returned-from-db-fn-without-id) (db.cljc:1253)
Uncaught exception, not in assertion.
	datascript.test.transact/fn--15645 	test/datascript/test/transact.cljc	:480
	datascript.conn/transact!          	src/datascript/conn.cljc          	:111
	datascript.conn/transact!          	src/datascript/conn.cljc          	:115
	datascript.conn/-transact!         	src/datascript/conn.cljc          	:86
	clojure.core/swap!                 	core.clj                          	:2369
	datascript.conn.Conn/swap          	conn.cljc                         	:11
	datascript.conn.Conn/swap*         	conn.cljc                         	:11
	clojure.core/apply                 	core.clj                          	:669
	datascript.conn/-transact!/fn--6013	src/datascript/conn.cljc          	:88
	datascript.conn/with               	src/datascript/conn.cljc          	:29
	datascript.db/transact-tx-data     	src/datascript/db.cljc            	:1923
	datascript.db/transact-tx-data-impl	src/datascript/db.cljc            	:1873
	datascript.db/transact-add         	src/datascript/db.cljc            	:1591
	datascript.db/entid-strict         	src/datascript/db.cljc            	:1261
	datascript.db/entid                	src/datascript/db.cljc            	:1253
ExceptionInfo: Expected number or lookup ref for entity id, got nil {:error :entity-id/syntax, :entity-id nil}
```

Cause bisected down to 3915250.

tonsky:
Fixed in 1.7.1

PR #475, opened 2024-06-20
Fix :db.fn/call returning entity without :db/id

DerGuteMoritz:
Potential fix for #474, not sure if ideal.


tonsky:
Yeah, it’s great. Thanks!

PR #477, opened 2024-07-20
fix: error when find-datom on empty-db

RCmerci:
`d/find-datom` will throw an error when calling on an empty-db.
Is this behavior by design or a mistake?

``` clojure
> (d/find-datom (d/empty-db {:age {:db/index true}}) :eavt)
Execution error (NullPointerException) at me.tonsky.persistent-sorted-set/seek (persistent_sorted_set.clj:51).
Cannot invoke "me.tonsky.persistent_sorted_set.Seq.seek(Object)" because "seq" is null
```

I expect it to return `nil` when an empty-db is used as arg to `find-datom`.


tonsky:
Thanks! Pushed 1.7.3

PR #479, opened 2024-10-13
Patch es6 iterators not having Symbol.iterator function

### LINKED ISSUE #249: JS Iterator protocol for entries
The README indicates that `entity` implements the ES6 Map interface. However it's not actually iterable because it doesn't implement the iterable protocol. To do so, an `entity` must also have the method `Symbol.iterator` as a computed method.

For example, using the node repl, you can test this:

```
e = d.entity(db, 1);
[...e.keys()] // TypeError: e.keys(...) is not iterable
[...e.values()] // TypeError: e.values(...) is not iterable
[...e] // TypeError: e is not iterable
```

Basically this function is missing:

```
e[Symbol.iterator]
```

And for each of the objects returned by `keys`, or `values` or `entries`.

kasbah:
closes #249

This should probably be fixed in Clojurescript itself but I'm out of my depth here already (I am a Clojure noob).

We can't do `(es6-iterable ES6Iterator)` because that would be infinite recursion so we patch the `ES6Iterator` prototype with the `Symbol.iterator` directly, sans recursion.

Interestingly enough `cljs.core/keys` already returns a `KeySeq` which has `Symbol.iterator` but no `next` function. Applying `es6-iterator`, as we do, seems give it the `next` function but strip `Symbol.iterator`.

This doesn't yet fix `[...e]` because I'm not quite sure what the intention there is, I guess the same as `[...e.entries()]`?

### DIFF
--- a/src/datascript/impl/entity.cljc
+++ b/src/datascript/impl/entity.cljc
@@ -45,6 +45,16 @@
          [a (multival->js v)]
          [a v]))))
 
+#?(:cljs
+  (unchecked-set (.-prototype ES6Iterator) cljs.core/ITER_SYMBOL
+     (fn []
+       (this-as this# this#))))
+
+#?(:cljs
+  (unchecked-set (.-prototype ES6EntriesIterator) cljs.core/ITER_SYMBOL
+     (fn []
+       (this-as this# this#))))
+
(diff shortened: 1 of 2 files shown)


tonsky:
Awesome! Thank you!

PR #484, opened 2025-03-06
Feat/timeout for `q` & `pull`

baibhavbista:
Adds a timeout to `q` and `pull`

Mostly inspired from the timeout approach in datalevin: https://github.com/juji-io/datalevin/blob/master/src/datalevin/timeout.clj

Is this approach correct, @tonsky ? (seems to work, based on my tests)

### DIFF
--- a/src/datascript/query.cljc
+++ b/src/datascript/query.cljc
@@ -13,6 +13,7 @@
                                                  FindColl FindRel FindScalar FindTuple PlainSymbol
                                                  RulesVar SrcVar Variable]])]
     [datascript.pull-api :as dpa]
+    [datascript.timeout :as timeout]
     [datascript.util :as util])
   #?(:clj
      (:import
@@ -48,6 +49,10 @@
      (.write w ", :tuples [")
      (.write w (str/join " " (map seq (:tuples r))))
      (.write w "]}")))
+
+(defn relation! [attrs tuples]
+  (timeout/assert-time-left)
+  (Relation. attrs tuples))
 
 
 ;; Utilities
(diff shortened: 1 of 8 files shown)


src/datascript/query.cljc
```diff
@@ -975,38 +983,41 @@
...
+    (->> (for [tuple resultset]
+           (mapv
+            (fn [parsed-opts el]
+              (if parsed-opts
+                (dpa/pull-impl parsed-opts el)
+                el))
+            resolved
+            tuple))
+         ;; realize lazy seq because this is the last step anyways, and because if we don't realize right now then binding for timeout/*deadline* does not work
+         doall)))
```
baibhavbista, line 995: another change which is not present in [datalevin timeout PR](https://github.com/juji-io/datalevin/pull/150). This is needed otherwise `pull` within `find` spec of `q` do not obey timeout

src/datascript/query.cljc
```diff
@@ -975,38 +983,41 @@
...
-      (:with q)
-      (mapv #(vec (subvec % 0 result-arity)))
-      (some dp/aggregate? find-elements)
-      (aggregate find-elements context)
-      (some dp/pull? find-elements)
-      (pull find-elements context)
-      true
-      (-post-process find (:qreturn-map parsed-q)))))
+  (let [parsed-q      (lru/-get *query-cache* q #(dp/parse-query q))]
+    (binding [timeout/*deadline* (timeout/to-deadline (:qtimeout parsed-q))]
```
baibhavbista, line 999: main change here is just the addition of the `binding` in the middle of the `let`

baibhavbista:
Sorry just seeing your reply now @tonsky !

The main usecase is when allowing end users to be able to run their own queries. Queries not properly written can cause combinatorial explosion of candidates, and lead either to OOM or to the application freezing for a non-allowably-long period of time (common use case of datascript running in the main thread in browser). In these cases, a timeout allows one to stop the query partway through if the time taken exceeds a value we deem reasonable.

Re: your implementation notes, I will go through them properly next week. BTW, as I mentioned in the PR description, most of your "why"s could be answered by "it was done that way in datalevin, and the relevant parts of the two codebases looked similar enough that I just recreated the same changes here in datascript". I will more properly think about your notes and reply/fix soon though

Issue #424, opened 2022-02-07
Allow calling `clojure.core` functions in queries


zoren:
In Datomic:
```
(q
 '[:where
   [(ground {}) ?m]
   [(assoc ?m :x 5) ?m2]
   :find ?m2 .])
```
returns `{:x 5}`. But it fails in datascript with `clojure.lang.ExceptionInfo: Unknown function 'assoc in [(assoc ?m :x 5) ?m2]`

Should we allow all functions in `clojure.core` (except `eval`) in datascript as well?

tonsky:
I think you can call any function by fully qualifying its name (in Clojure). We might also check clojure.core by default, I guess

zoren:
> I think you can call any function by fully qualifying its name (in Clojure). We might also check clojure.core by default, I guess

Ah thanks didn't know. If we did we wouldn't need to list them explicitly here: https://github.com/tonsky/datascript/blob/master/src/datascript/built_ins.cljc#L80

tonsky:
Those are for ClojureScript, since it doesn’t have runtime resolve

PR #488, opened 2025-06-23
Document namespaced clojure function requirement

jessie-ross:
I haven't used Datomic, but I have been following it's documentation to learn Datascript, I believe have found this gap in the behaviour between Datomic and Datascript queries, it would be good to document it:
- https://github.com/tonsky/datascript/issues/424

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,7 @@ Interface differences:
 * Attributes do not have to be declared in advance. Put them to schema only when you need special behaviour from them
 * Any type can be used for values
 * No `:db/ident` for attributes, keywords are _literally_ attribute values, no integer id behind them
+* All Clojure functions must be fully namespaced in queries
 * No schema migrations
 * No full-text search, no partitions
 * No external dependencies


tonsky:
This is not exactly true. These functions don’t need to be qualified: https://github.com/tonsky/datascript/blob/8c01d1b7e401ab1765f2933164267922f550d753/src/datascript/built_ins.cljc#L81-L99

For custom functions, in CLJS even fully-qualified won’t work. You’ll have to pass fn as an argument.

Finally, Datomic seems to require fully-qualified too:

Screenshot 2025-06-23 at 11 59 32

jessie-ross:
> Finally, Datomic seems to require fully-qualified too:
 
Ah I wasn't clear enough perhaps, I mean `clojure.core` functions, see this example:

```clojure
(require '[datomic.api :as d :refer [q db]])

(def uri "datomic:dev://localhost:4334/mbrainz-1968-1973")
(def conn (d/connect uri))
(def the-db (d/db conn))


;; Check I am running this correctly:
(d/q '[:find ?artist-name .
       :in $
       :where
       [?a :artist/name   ?artist-name]
       [?t :track/artists ?a]
       [?t :track/name    "Baby's Heartbeat"]]
     the-db) ;; => "John Lennon"

;; clojure.core/concat is included
(d/q '[:find ?result .
       :in $ ?a ?b
       :where
       [(concat ?a ?b) ?result]]
     the-db ["a"] ["b"]) ;; => ("a" "b")

;; Checking if it includes symbols present in the current namespace
(defn foo [a] (str a "-foo"))

;; It doesn't
(d/q '[:find ?result .
       :in $ ?a
       :where
       [(foo ?a) ?result]]
     the-db "a") ;; => Unable to resolve foo in this context

(d/q '[:find ?result .
       :in $ ?a
       :where
       [(user/foo ?a) ?result]]
     the-db "a") ;; => "a-foo"
```

I was following [this blog post](https://hashrocket.com/blog/posts/using-datomic-as-a-graph-database) and tripped over the difference between datomic and datascript on functions like `concat`, and `butlast`. I am not sure whether just Datomic has a larger amount of functions allowed or whether it includes `clojure.core`.

Now that I've solved the issue it isn't of great importance to me, but I was thinking it might be nice to document it.

PR #492, opened 2025-10-09
perf: Skip storage writes for no-op transactions

jxonas:
### Summary

Optimizes storage behavior by skipping persistence when transactions produce no datoms. This avoids unnecessary I/O for empty transactions and operations that add redundant or retract non-existent facts.

### Problem

Currently, when a connection is associated with `:storage`, `d/transact!` always calls `storage/-store`, even when the transaction's `:tx-data` is empty. This happens for:

- Empty transactions: `(d/transact! conn [])`
- Adding redundant facts
- Retracting non-existent facts

These no-op transactions trigger unnecessary storage writes, impacting performance in scenarios with many such transactions.

### Solution

Modified `-transact!` in `src/datascript/conn.cljc` to check `(seq datoms)` before executing storage logic. Storage writes are now skipped when `:tx-data` is empty.

**Key points:**
- No change to transaction semantics
- TxReport structure remains unchanged
- Listeners still fire normally
- Tempid resolution unaffected
- All existing tests pass

### DIFF
--- a/src/datascript/conn.cljc
+++ b/src/datascript/conn.cljc
@@ -89,21 +89,22 @@
           (vreset! *report r)
           (:db-after r))))
     #?(:clj
-       (when-some [storage (storage/storage @conn)]
-         (let [{db     :db-after
-                datoms :tx-data} @*report
-               settings (set/settings (:eavt db))
-               *atom    (:atom conn)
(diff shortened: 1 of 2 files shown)


tonsky:
Thanks! Merged and released 1.7.7

tonsky:
\*1.7.8

PR #494, opened 2026-02-16
Migrate persistent-sorted-set to org.replikativ fork

BuddhiLW:
## Summary

- Migrate from `persistent-sorted-set/persistent-sorted-set 0.3.0` to `org.replikativ/persistent-sorted-set 0.3.118`
- Update both `deps.edn` and `project.clj`

## Motivation

The original `persistent-sorted-set/persistent-sorted-set` artifact appears to be unmaintained (last release: 0.3.0). The `org.replikativ/persistent-sorted-set` fork is actively maintained and version 0.3.118 preserves the same `me.tonsky.persistent-sorted-set` Clojure namespace — fully API-compatible, no code changes needed.

This resolves classpath conflicts in projects that combine DataScript with [Datahike](https://github.com/replikativ/datahike), which already depends on the replikativ fork. Currently, both artifacts end up on the classpath with different group IDs for the same library, causing version resolution issues (especially when `org.replikativ/persistent-sorted-set` resolves to 0.4.x which renamed the namespace to `org.replikativ.persistent-sorted-set`).

## Testing

- Verified `datascript.core` loads successfully with the new dependency
- No code changes required — the fork preserves the `me.tonsky.persistent-sorted-set` namespace at 0.3.x


tonsky:
Hey, thanks for opening the issue. I spoke with @whilo about the fork, for me it’s too much work to merge or even switch to it. If it’s as simple as changing a dependency then you can do it in your own deps.edn with :exclude, but if it’s not I am not prepared at the moment to do the migration. So we decided that they will keep working on their fork as a separate project (which is I think for the better: it seems far ahead of PSS).

Namespace wise, I think two different libraries should provide two different namespaces, but they should change it IMO

